PR #1, opened 2023-11-18
tamagui configuration - colors

0xLe0x:
- added `#C3AB8E` as $Primary color
- moved the brand colors from Uniswap
```
  white: '#FFFFFF',
  black: '#000000',
  gray50: '#F5F6FC',
  gray100: '#E8ECFB',
  gray150: '#D2D9EE',
  gray200: '#B8C0DC',
  gray250: '#A6AFCA',
  gray300: '#98A1C0',
  ...
```


packages/ui/src/themes/colors.ts
```diff
@@ -0,0 +1,173 @@
+/**
+ * change this export to update the colors
+ *
+ * 1. './colors-default' - the default tamagui theme
+ * 2. './colors-neon' - based on the default theme, less saturated
```
0xBigBoss, line 5: we can drop extra colors here as well.

packages/ui/src/themes/colors.ts
```diff
@@ -0,0 +1,173 @@
...
+ * change this export to update the colors
+ *
+ * 1. './colors-default' - the default tamagui theme
+ * 2. './colors-neon' - based on the default theme, less saturated
+ * 2. './colors-pastel' - based on the default theme, more saturated
+ */
+
+export * from './colors-default'
+
+export const brandColors = {
```
0xBigBoss, line 11: so the big idea here is that we can use these as tamagui tokens in our component props? `backgroundColor="$primary"`?

does it make sense to add it as a child theme like we do for the other colors? That way we can just wrap our components in a `<Theme name="brand">{children}</Theme>` or set it at the component level as well, `theme="brand"`.

https://github.com/0xsend/sendapp/blob/a1b310a5b97d80fb5ee734090cbae14e988e38ab/packages/ui/src/themes/theme.ts#L45-L54

packages/ui/src/themes/token-colors.ts
```diff
@@ -0,0 +1,96 @@
...
+export {
+  darkColor,
+  darkPalette,
+  darkTransparent,
+  lightColor,
+  lightPalette,
+  lightTransparent,
+} from './colors'
+
+export const colorTokens = {
```
0xBigBoss, line 34: do we need to include brand colors here too?

0xBigBoss (approved):
Just a few nits and questions. Once we have the questions answered and the cleanups included. Feel free to merge @0xLe0x

packages/ui/src/themes/colors-neon.ts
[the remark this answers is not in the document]
0xLe0x: removed

packages/ui/src/themes/colors-pastel.ts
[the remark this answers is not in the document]
0xLe0x: removed

packages/ui/src/themes/colors.ts
0xLe0x: removed

packages/ui/src/themes/colors.ts
0xLe0x: The newly added colors are tokens yet, not sure if it's possible to add it as a child theme.

packages/ui/src/themes/token-colors.ts
0xLe0x: yes, it's just imported to be included in `color` that is used to `createTokens` in tamagui.config.ts

PR #6, opened 2023-12-03
fix formatNumpadInput

0xLe0x:


0xBigBoss (approved):
Good catch, thanks @0xLe0x

PR #16, opened 2023-12-08
send modals to screens

0xLe0x:
- turn modals into screens
- add confirm modal
- fix responsiveness


0xBigBoss (changes requested):
I wouldn't make them standalone pages, it would be a nightmare to keep the state persisted between them. A single send page is the way to do it and it should swap between the screens like u were doing it before when they were modals.

To keep the state between the different components I would create a new react context/or a mobx state tree and share it with the components.

0xLe0x:
- used `framer-motion` library for animation between screens
- can use `react-navigation` on native
- created `AnimationLayout` for reuse

apps/next/pages/send.tsx
```diff
@@ -1,4 +1,4 @@
-import { SendScreen } from 'app/features/send/screen'
+import { SendPage } from 'app/features/send/screens/send'
```
0xBigBoss, line 1: feel like we should stick to screens in this folder. wdyt? 

pages won't be 1-1 on native.

packages/app/components/layout/animation-layout.tsx
```diff
@@ -0,0 +1,28 @@
...
+  } as const,
+})
+
+export const AnimationLayout = ({ children, currentKey, direction }: AnimationLayoutProps) => {
+  const enterVariant = direction === 1 || direction === 0 ? 'isRight' : 'isLeft'
+  const exitVariant = direction === 1 ? 'isLeft' : 'isRight'
+
+  return (
+    <AnimatePresence enterVariant={enterVariant} exitVariant={exitVariant}>
+      <YStackEnterable key={currentKey} animation="quick" fullscreen x={0} opacity={1}>
```
0xBigBoss, line 23: this animation is pretty slick, but also not a fan of how it shakes at the end. it would be nice to have it just ease and then snap.

packages/app/features/send/components/send/confirm-modal.tsx
```diff
@@ -29,24 +30,32 @@ const CustomInput = styled(Input, {
   height: '$4.5'
 })
 
-export const SendItModal = ({ sendAmount, asset, tag, showModal, setCurrentModal }: SendItModalProps) => {
+const tag = {
```
0xBigBoss, line 33: I think maybe in the next PR, we create a TransferProvider or something similar to encapsulate this state and share it with the components. wdyt?

0xBigBoss (approved):
looks pretty good, I would say it's good enough to merge. We can probably start cleaning up the send amount input validation since there are a few bugs around that as far as UX (e.g. inputing something greater than your balance, sometimes pressing numbers doesn't do anything, button mashing breaks the input). Ideally we'd have unit tests for these things or even start an e2e test.

https://github.com/0xsend/sendapp/assets/95193764/7352bb04-d070-4c37-94f0-28882afbd54a

Or we can start a new screen. I'll leave it up to you.

0xLe0x:
> looks pretty good, I would say it's good enough to merge. We can probably start cleaning up the send amount input validation since there are a few bugs around that as far as UX (e.g. inputing something greater than your balance, sometimes pressing numbers doesn't do anything, button mashing breaks the input). Ideally we'd have unit tests for these things or even start an e2e test.
> 
>  Screen.Recording.2023-12-11.at.13.30.20.mov 
> Or we can start a new screen. I'll leave it up to you.

`sometimes pressing numbers doesn't do anything` - Currently it does not handle input when fractional part is 3 digits or longer. Do we need to input more than 3? if yes, i'll eliminate that restriction

0xBigBoss:
> > looks pretty good, I would say it's good enough to merge. We can probably start cleaning up the send amount input validation since there are a few bugs around that as far as UX (e.g. inputing something greater than your balance, sometimes pressing numbers doesn't do anything, button mashing breaks the input). Ideally we'd have unit tests for these things or even start an e2e test.
> > Screen.Recording.2023-12-11.at.13.30.20.mov
> > Or we can start a new screen. I'll leave it up to you.
> 
> @0xBigBoss `sometimes pressing numbers doesn't do anything` - Currently it does not handle input when fractional part is 3 digits or longer. Do we need to input more than 3? if yes, i'll eliminate that restriction

Hmm, lets just remove for now. Later on, I think it will be currency dependent. e.g. USDC/USDT probably will only need 2 decimals. ETH will probably need at least 6. but let's push that for another day.

To me, there should be no user input without feedback. I guess it's just surprising to the user, we should show some UI feedback by disabling the decimal button or some other appropriate thing.

PR #26, opened 2023-12-22
Add Anvil Node contribution docs

youngkidwarrior:
Add some instructions for running anvil for mainnet and base on the expected dev ports


0xBigBoss (approved):
Good looking out. I would encourage you to try tilt since that's what I use to orchestrate the CI. https://github.com/0xsend/sendapp/blob/main/Tiltfile

The syntax is a subset of python called starlark, but it's very easy to read and see how it all works together.

PR #39, opened 2023-12-22
QR Screens

0xLe0x:


packages/app/features/send/screens/qrscan/qr-amount.tsx
```diff
@@ -0,0 +1,107 @@
...
+  YStack,
+} from '@my/ui'
+import { IconClose } from 'app/components/icons'
+import { IQRScreenProps } from 'app/features/send/types'
+import { NumPad } from 'app/features/send/components/numpad'
+import { RequestConfirmModal, SendConfirmModal } from 'app/features/send/components/modal'
+import { useTransferContext } from 'app/features/send/providers/transfer-provider'
+
+export const QRAmountScreen = ({ setCurrentScreen, sendOrRequest }: IQRScreenProps) => {
+  const { transferState, updateTransferContext } = useTransferContext()
```
0xBigBoss, line 17: @0xLe0x sorry I missed this before, but instead of having one giant object for the context, it should be built up using different `useState` calls. That way every component that depends on this context won't re-render. This is a really complicated example, but something like this is preferable https://github.com/0xsend/superhack/blob/32657b564d4e49c83e5859e274a8ca21e516612c/packages/app/provider/WalletProvider.tsx#L9

packages/app/features/send/screens/qrscan/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  'qr-amount': QRAmountScreen,
+  'qr-share': QRShareScreen,
+};
+
+export const QRScreen = () => {
+  const [[currentScreen, direction, sendOrRequest], setCurrentScreenState] = useState<
+    [QRScreenType, number, 'Send' | 'Request' | undefined]
+  >(['qr-scan', -1, undefined]);
+
+  const setCurrentScreen = (
```
0xBigBoss, line 23: I really like this pattern you've created to do sub-screen navigation. Think about renaming it to be more obvious what it's doing like SubScreen or just CurrentComponent. 

However, I think we should avoid using props at all costs unless they are for style passing or a single object. Let's move this `currentScreen` pattern into the screen's context similar to what I asked for the transfer context. Then rinse and repeat for the other screens. 

Then, add the currentScreen state to that context. Then if a sub-component wants to request a navigation they should do something like `useCurrentScreen` and then `setCurrentScreen` to navigate. 

Also make the direction a type

```ts
export type ANIMTE_DIRECTION_LEFT = -1 as const;
export type ANIMATE_DIRECTION_RIGHT = 1 as const;
export type ANIMATE_DIRECTION = ANIMTE_DIRECTION_LEFT | ANIMATE_DIRECTION_RIGHT
```

packages/app/features/send/types.ts
```diff
@@ -16,6 +16,24 @@ export interface IConfirmModalProps {
   setShowModal: (showModal: boolean) => void
 }
 
+export interface ISendRequestModalProps extends IQRScreenProps {
+  showModal: boolean
+  setShowModal: (showModal: boolean) => void
+  to?: ITag
```
0xBigBoss, line 22: I think it's time  I give you a demo of the backend, we already have a lot of types for the API

https://github.com/0xsend/sendapp/blob/88bdab891eea1d514b6aa47cd387d42b216c6030/supabase/database-generated.types.ts#L399-L404

packages/app/features/send/components/modal/profile-modal.tsx
```diff
@@ -0,0 +1,183 @@
...
+  tag,
+  setCurrentScreen,
+}: IProfileModalProps) => {
+  const [showProfileQRModal, setShowProfileQRModal] = useState(false)
+
+  const { resolvedTheme } = useThemeSetting()
+
+  return (
+    <>
+      <Theme name={resolvedTheme as ThemeName}>
```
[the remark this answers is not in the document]
0xLe0x: It's only for modals, not sure why Theme doesn't work in modals unless `<Theme name={resolvedTheme as ThemeName} />` isn't added.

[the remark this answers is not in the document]
0xBigBoss: Ahh I think I saw this before when building the tag checkout. I believe it’s due to the modal being rendered outside of the providers in a portal. Unsure if it’s a bug or a feature with tamagui modals.

packages/app/features/send/screens/qrscan/index.tsx
0xLe0x: ❤️

packages/app/features/send/screens/receive/receive-tag.tsx
```diff
@@ -12,9 +12,9 @@ import {
   styled
 } from "@my/ui"
 import { Link } from '@my/ui/src/components'
```
0xBigBoss, line 14: just import from the @my/ui package

packages/app/features/send/types.ts
```diff
@@ -27,19 +46,16 @@ export interface ITag {
   avatar: string
 }
 
-export type SendScreenType = 'send' | 'send-tag' | 'send-it'
-
-export type ReceiveScreenType = 'receive-qrcode' | 'receive-tag' | 'receive-amount'
+export type SendScreenType = 'home' | 'send' | 'send-tag' | 'send-it'
```
0xBigBoss, line 50: I usually prefer something like this so you can easily see the references without having to search.

```ts
export enum SendScreen {
  HOME = 'home',
  SEND = 'send',
  SEND_TAG = 'send-tag',
  SEND_IT = 'send-it'
}

export type SendScreenType = SendScreen.HOME | SendScreen.SEND | SendScreen.SEND_TAG | SendScreen.SEND_IT
```

0xBigBoss (commented):
Lookin' really good! I say let's merge after the few comments I had.


P.S. i think once we get you up to speed on the backend, we need to consider how to integrate these pages into our solito routers.

PR #40, opened 2023-12-22
next: fix reset css

0xBigBoss:


youngkidwarrior (commented):
This solution might be a bit less hacky. Not sure if it works for our case


```
    width: 100vw;
    height: 100vh;
    max-width:100%;
    ```

https://stackoverflow.com/a/23367686

**Review by 0xBigBoss — COMMENTED** [review]
I just fat fingered this solution. not sure how I messed this one up. https://stackoverflow.com/questions/33606565/is-it-possible-to-calculate-the-viewport-width-vw-without-scrollbar/34884924#34884924

> This solution might be a bit less hacky. Not sure if it works for our case


```
    width: 100vw;
    height: 100vh;
    max-width:100%;
```

https://stackoverflow.com/a/23367686

PR #41, opened 2023-12-23
Port old header and update style

youngkidwarrior:
Create a home header component that matches the [new style from figma](https://www.figma.com/file/k7ct2aqK1zUvZripOWtW8h/send-landing?type=design&node-id=1167-8450&mode=design&t=cTIAsAOgjt0Xw6tQ-4)

Not sure whether this was supposed to be wagmi connector or filled from backend. The current app doesn't allow the user to change wallets. Do we plan to have that feature?
Screenshot 2023-12-22 at 11 47 18 PM

PR #28, opened 2023-12-22
Implement Sidebar Component

youngkidwarrior:
This is a WIP for the [Sidebar Component](https://www.figma.com/file/k7ct2aqK1zUvZripOWtW8h/send-landing?type=design&node-id=1167-8121&mode=design&t=AePTMchMhjpGZyKl-0)

Notably Missing Functionality:

 - ~~No mobile support/sandwich icon~~
 - ~~NavLink Icons must be imported from figma~~
 - Radial Gradients are not implemented in tamagui
 - ~~Distribution page has weird sidescroll~~
 - ~~SendLogoSmall Component has no support for custom colors~~
 - No tests
 - No support for sticky position in tamagui. Need a refactor to work around this

PR #50, opened 2024-01-04
Home Bottom Sheet for mobile

youngkidwarrior:
Adds logic for bottom sheet on mobile. Uses url state instead of context

Not 100% happy with the naming of everything, probably will take some time to rethink

This PR depends on #28 and #41

### DIFF
--- a/packages/app/components/HomeHeader.tsx
+++ b/packages/app/components/HomeHeader.tsx
@@ -5,8 +5,18 @@ import { getReferralHref } from "app/utils/getReferralLink";
 import { IconClose, IconHamburger, IconSettings } from "./icons";
 import { useAccount, useConnect, useDisconnect } from "wagmi";
 import { IconStar } from "./icons/IconStar";
+import { useParams } from "app/routers/params";
 
 export function HomeHeader({ children }: { children: string }) {
+  const { params, setParams } = useParams()
+  const handleHomeBottomSheet = () => {
+    setParams({
+      // params will merge with existing ones
+      nav: params.nav ? undefined : "home",
+    })
+  }
+
+
   return (
     <Header w="100%" $gtSm={{ miw: 804 }}>
       <XStack jc="space-between" fd="row" ai="center">
@@ -18,7 +28,7 @@ export function HomeHeader({ children }: { children: string }) {
           <WagmiAccountInfo />
         </XStack>
         <XStack $gtLg={{ display: "none" }} ai="center" space="$2" height="$4">
-          <Button bg="transparent" icon={<IconHamburger size={"$3"} />} />
+          <Button onPress={handleHomeBottomSheet} bg="transparent" icon={<IconHamburger size={"$3"} />} />
         </XStack>
       </XStack>
     </Header >)
--- a/packages/app/components/HomeHeader.tsx
+++ b/packages/app/components/HomeHeader.tsx
@@ -5,15 +5,12 @@ import { getReferralHref } from "app/utils/getReferralLink";
 import { IconClose, IconHamburger, IconSettings } from "./icons";
 import { useAccount, useConnect, useDisconnect } from "wagmi";
 import { IconStar } from "./icons/IconStar";
-import { useParams } from "app/routers/params";
+import { useNav } from "app/routers/params";
 
 export function HomeHeader({ children }: { children: string }) {
-  const { params, setParams } = useParams()
+  const [nav, setNavParam] = useNav()
   const handleHomeBottomSheet = () => {
-    setParams({
-      // params will merge with existing ones
-      nav: params.nav ? undefined : "home",
-    })
+    setNavParam(nav ? undefined : "home", { webBehavior: "replace" })
   }
 
 
--- a/packages/app/components/sidebar/HomeSideBar.tsx
+++ b/packages/app/components/sidebar/HomeSideBar.tsx
@@ -7,7 +7,7 @@ import { SideBarFooterLink } from "app/components/sidebar/SideBarFooterLink";
 import { twitter as twitterSocial, telegram as telegramSocial } from 'app/data/socialLinks'
 import { usePathname } from "app/utils/usePathname";
 
-import { useParams } from "app/routers/params";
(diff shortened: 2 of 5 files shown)


0xBigBoss (commented):
Is the idea that we may have multiple navs? Else I am kinda confused by this implementation. 

Isn't there a way to watch for path changes in solito? Couldn't we just turn this prop into component state then?  I feel like having each link set the param is kinda an antipattern and isn't required.

other than that, this PR looks spicy and really close.

youngkidwarrior:
@0xBigBoss the original version of this PR used context to achieve the open and close functionality.

The nice thing about using query params is that 

1. It avoids context/global state
2. It allows linking with the bottom sheet open
3. We can handle multiple navbar states in one place by `switch`ing over the allowed `nav` states
  I think in another PR you had told me there were going to be subroutes with different navbars.


Where I think this falls short is that the state of the navbar can be adjusted via the url, so we will need some sort of gating logic at the component level if there are any navbars that require auth

packages/app/components/HomeHeader.tsx
```diff
@@ -5,8 +5,18 @@ import { getReferralHref } from "app/utils/getReferralLink";
 import { IconClose, IconHamburger, IconSettings } from "./icons";
 import { useAccount, useConnect, useDisconnect } from "wagmi";
 import { IconStar } from "./icons/IconStar";
+import { useParams } from "app/routers/params";
 
 export function HomeHeader({ children }: { children: string }) {
+  const { params, setParams } = useParams()
+  const handleHomeBottomSheet = () => {
```
0xBigBoss, line 12: I see now, let's rename to something more obvious. I didn't get it at first. `openHomeNav` or something

packages/app/components/sidebar/HomeSideBar.tsx
```diff
@@ -52,9 +63,9 @@ const HomeBottomSheet = () => {
       </Link>
       <Nav display="flex" flex={2} justifyContent={"center"} alignItems="center">
         <YStack gap={"$4"} alignItems='flex-start' justifyContent='center'>
-          <SideBarNavLink icon={<IconDashboard size={"$2"} />} text={"Dashboard"} href={"/"} isActive={pathName === "/"} />
-          <SideBarNavLink icon={<IconDistributions size={"$2"} />} text={"Distributions"} href={"/distributions"} isActive={pathName.includes("/distributions")} />
-          <SideBarNavLink icon={<IconSLogo size={"$2"} />} text={"Leaderboard"} href={"/leaderboard"} isActive={pathName.includes("/leaderboard")} disabled={true} hoverStyle={{ cursor: "not-allowed" }} />
+          <SideBarNavLink onPress={handleHomeBottomSheet} icon={<IconDashboard size={"$2"} />} text={"Dashboard"} href={"/"} isActive={pathName === "/"} />
```
0xBigBoss, line 66: Why do we need these `onPress`?

packages/app/components/sidebar/HomeSideBar.tsx
```diff
@@ -38,10 +40,19 @@ const HomeSideBar = ({ ...props }: YStackProps) => {
...
+    setParams({
+      // params will merge with existing ones
+      nav: params.nav ? undefined : 'home'
+    })
+  }
+
   const pathName = usePathname()
   return (
-    <BottomSheet open>
+    <BottomSheet open={open} onOpenChange={handleHomeBottomSheet}>
```
0xBigBoss, line 55: Also, what's the purpose of the `onOpenChange` here?

packages/app/components/sidebar/HomeSideBar.tsx
youngkidwarrior: Good catch. Actually we don’t need this. This was needed when using context in order to close it when link was pressed. now it will close automatically when we link to new page

packages/app/components/sidebar/HomeSideBar.tsx
youngkidwarrior: This plugs into the `Sheet.Overlay` and `Sheet.Handle` so when overlay is pressed or sheet is swiped away, it will also update the query parameter accordingly

packages/app/components/HomeHeader.tsx
youngkidwarrior: Exactly, not happy with the names, but wanted to get it out to you to get opinion on query params or context

youngkidwarrior: It’s also interesting that with this solution, a page could show a different nav depending on what the user is doing. 

Might be some interesting UX opportunities

packages/app/components/sidebar/HomeSideBar.tsx
```diff
@@ -70,7 +81,8 @@ const HomeBottomSheet = () => {
 
 export const HomeSideBarWrapper = ({ children }: { children?: React.ReactNode }) => {
   const media = useMedia()
-  if (media.gtMd)
+  const { params } = useParams()
```
0xBigBoss, line 84: move this prop into the sheet as state.

packages/app/components/HomeHeader.tsx
0xBigBoss: Now, that I reviewed the solito API, i think the names are fine. I like the API you defined with managing all the navs with one param.

packages/app/components/sidebar/HomeSideBar.tsx
0xBigBoss: I see, let's make it more clear. In my opinion something like this is cleaner and more obvious. 

```ts
const HomeBottomSheet = () => {
  const [nav, setNavParam] = useParam('nav')

  const onOpenChange = (open: boolean) => {
    if (open) setNavParam('home')
    else setNavParam(undefined)
  }

  const pathName = usePathname()
  return (
    <BottomSheet open={nav === 'home'} onOpenChange={onOpenChange}>
    //...
    </BottomSheet>
  )
}
```

That way you don't need to pass any props to this component. On native, we can watch the pathname and if it changes, just close the sheet by changing the nav param to undefined.

PR #59, opened 2024-01-10
Add User referrals policy and user referrals count

youngkidwarrior:


supabase/migrations/20240110232159_allow_referrer_to_read_referrals.sql
```diff
@@ -0,0 +1,20 @@
+set check_function_bodies = off;
+
+CREATE OR REPLACE FUNCTION public.user_referrals_count()
```
0xBigBoss, line 3: This is pretty close, but instead it needs security_definer which would inherit the permissions of the function creator (e.g. root).  You can see an example here and how we confirm tags after users have paid. https://github.com/0xsend/sendapp/blob/4cf160540493a9dd42e84fd9557b53b9b33ff537/supabase/migrations/20230829195803_update_confirm_tags_all_are_paid.sql#L10-L16. However, that function also has protections to ensure only the service_role can execute it.

That way we can disable the RLS policy below. Since enabling that policy allows users to view all of their referrals and the associated tags.

You can see how supabase will allow users to fetch the referrals.
```ts
 const { data, error } = await supabase.from('referrals').select('*')
console.log(data, error)
```
Screenshot 2024-01-10 at 21 07 48

supabase/database-generated.types.ts
0xBigBoss, line 1: Thanks for turning on the formatter. I think I missed this file. Could you ignore it in the biome.json?

https://github.com/0xsend/sendapp/blob/ykw/add-user-referrals-policy-and-user-referrals-count/biome.json#L20-L21
```json
  "formatter": {
    "indentStyle": "space",
    "indentWidth": 2,
    "lineWidth": 100,
    "lineEnding": "lf",
    "ignore": ["package.json", "**/generated.ts", "**/database-generated.types.ts"]
  }
```

packages/app/utils/useUser.ts
```diff
@@ -62,13 +62,13 @@ export const useUser = () => {
   })()
 
   const {
```
0xBigBoss, line 64: I thikn I made a mistake when I added this query here, I think we can move it into a new hook, `useReferralsCount` instead of fetching on every page.

supabase/migrations/20240110232159_allow_referrer_to_read_referrals.sql
youngkidwarrior: Do we not want a user to have access to all their referrals (but not other users referrals)?

Then they could verify who they have referred and who they haven't

0xBigBoss: We do, but not the raw data. eventually when we add privacy settings, perhaps we will let them view the tags they referred, but definitely not the user IDs.

youngkidwarrior:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/59?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

Current dependencies on/for this PR:
* `main`
  * **PR #41** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/41?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #59** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/59?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/0xsend/sendapp/59?utm_source=stack-comment).

supabase/migrations/20240110232159_allow_referrer_to_read_referrals.sql
```diff
@@ -0,0 +1,11 @@
+set check_function_bodies = off;
+
+CREATE OR REPLACE FUNCTION public.user_referrals_count(user_id uuid)
+ RETURNS integer
+ LANGUAGE plpgsql SECURITY DEFINER
+SET search_path TO 'public' AS $function$
+begin
+return (select count(*) from referrals
+where referrer_id=user_id);
```
0xBigBoss, line 9: If a users knows the ID of another user, they can look up the referral counts using this function, not really a big deal but something to know when creating functions using `SECURITY DEFINER` since the SQL effectively runs as root bypassing the RLS.

I would have dropped the user_id param and added the following SQL.

```sql
where referrer_id=auth.uid()
```

That way it is hard-coded into the function.

Another way could have been to create a view like I did here https://github.com/0xsend/sendapp/blob/ykw/add-user-referrals-policy-and-user-referrals-count/supabase/migrations/20230910014335_create_distribution_table.sql#L89

packages/api/src/routers/referrals.ts
```diff
@@ -0,0 +1,17 @@
+import { TRPCError } from '@trpc/server'
+import { supabaseAdmin } from 'app/utils/supabase/admin'
+import { z } from 'zod'
+import { createTRPCRouter, protectedProcedure } from '../trpc'
+
+export const referralsRouter = createTRPCRouter({
+  userReferralsCount: protectedProcedure.input(z.string()).query(async ({ input: userId }) => {
+    const { data: results, error } = await supabaseAdmin
```
0xBigBoss, line 8: is this route required? I thought you may be able to call this with regular user credentials like you were before.

I don't think admin is needed.

0xBigBoss (approved):
Gonna approve, but please see my comment about RLS and the function permissions. Up to u if you want to clean that up.

packages/api/src/routers/referrals.ts
youngkidwarrior: I'm pretty unfamiliar with Postgres/supabase. Apologies, I figured that by adding the security to `definer`, it would only be callable by admin.

supabase/migrations/20240110232159_allow_referrer_to_read_referrals.sql
youngkidwarrior: This is how I originally had it. Misunderstood what `security definer` means

0xBigBoss: Right, but because RLS was active and policies were enabled to select the rows, it would leak the raw data.

I proposed security definer since the function would run the queries as the owner(postgres) of the function which means it would bypass RLS. It would still require `auth.uid() = referrer_id`. This the relevant docs, https://arc.net/l/quote/rsuztfwi.

That way we can still call the function from using the supabase client with regular user credentials.

For the view approach I linked above it's a very similar approach. Views by default run their queries by the owner of the view, in this case, postgres. That means it bypasses RLS, which is why it has security_barrier on to ensure that it won't leak rows that an user role shouldn't have access to. See https://arc.net/l/quote/ocslvvni and https://arc.net/l/quote/cjosifld.

PR #88, opened 2024-01-23
Features - Settings and Profile screens

syedehsans1:
- Designed settings screen
- Designed profile edit screen
- Supabase integrated for profile


packages/app/components/icons/IconBack.tsx
0xBigBoss, line 1: All of the icons are missing the props being passed down.

Like so.

```tsx
 <Svg
      color={color as ColorTokens | undefined}
      width={size}
      height={size}
```

0xBigBoss (commented):
Didnt' get a chance to review all of it.

PR #144, opened 2024-03-01
Leo/fix balance format

0xLe0x:


packages/app/features/home/TokenDetails.tsx
```diff
@@ -20,9 +20,14 @@ const TokenDetails = ({ tokenAddress }: { tokenAddress: `0x${string}` | undefine
     if (balance.isPending) {
       return <Spinner size={'small'} />
     }
+    if (balance?.data?.value === undefined) {
```
0xBigBoss, line 23: this is fine for now, but i'd rather check if there's an error and show it to the user.

then use a our `assert` util so we can know when things break.

packages/app/utils/useSendAccountBalances.ts
```diff
@@ -24,12 +26,30 @@ export const useSendAccountBalances = () => {
     })
   }
 
-  let totalBalance = 0n
+  useEffect(() => {
```
0xBigBoss, line 29: make another hook for this in another file

0xBigBoss: It should use react query

packages/app/features/home/TokenDetails.tsx
```diff
@@ -15,14 +16,19 @@ const TokenDetails = ({ tokenAddress }: { tokenAddress: `0x${string}` | undefine
     chainId: baseMainnet.id,
   })
 
-  // @TODO: add balance error check
   if (balance) {
+    assert(!balance?.error, balance?.error?.message)
```
0xBigBoss, line 20: sorry I didn't mean like this. we should show this error to the user if it exists. not throw an error which is what assert will do.

packages/app/utils/useSendAccountBalances.ts
```diff
@@ -24,12 +27,17 @@ export const useSendAccountBalances = () => {
     })
   }
 
-  let totalBalance = 0n
+  assert(!error, error?.message)
```
0xBigBoss, line 30: this will hard throw and we don't have any error boundaries setup.

packages/app/utils/useTokenPrices.ts
```diff
@@ -0,0 +1,19 @@
...
+    data: tokenPrices,
+    error,
+    isLoading,
+  } = useQuery({
+    queryKey: ['tokenPrices'],
+    queryFn: async () => {
+      const res = await fetch(
+        'https://api.coingecko.com/api/v3/simple/price?ids=ethereum,usd-coin,send-token&vs_currencies=usd'
+      )
+      return await res.json()
```
0xBigBoss, line 14: check if the response is ok first and that the prices are all set. We are using zod. please create a schema and parse it at runtime too.

packages/app/utils/useSendAccountBalances.ts
0xLe0x: just remove assert?

packages/app/features/home/TokenDetails.tsx
```diff
@@ -15,14 +17,22 @@ const TokenDetails = ({ tokenAddress }: { tokenAddress: `0x${string}` | undefine
     chainId: baseMainnet.id,
   })
 
-  // @TODO: add balance error check
   if (balance) {
+    if (balance.isError) {
+      toast.show(balance.error.message)
```
0xBigBoss, line 22: Toasts are not great for errors. We should show it inline instead of the balance. 

Something like this would totally be fine. 

Error occurred while fetching balance. ${err.message}

packages/app/utils/useSendAccountBalances.ts
0xBigBoss: Yeah. We use assert more for type assertions and things that should never happen. 

Errors will occur. In the UI, the components themselves should handle them accordingly by giving the user feedback on possible resolutions or at least telling them what went wrong so they can copy-paste it to show us.

packages/app/utils/useTokenPrices.ts
```diff
@@ -0,0 +1,37 @@
...
+    queryFn: async () => {
+      const res = await fetch(
+        'https://api.coingecko.com/api/v3/simple/price?ids=ethereum,usd-coin,send-token&vs_currencies=usd'
+      )
+      if (!res.ok) {
+        throw new Error(`Failed to fetch token prices. Status: ${res.status}`)
+      }
+
+      const json = await res.json()
+      TokenPricesSchema.parse(json)
```
0xBigBoss, line 30: Return the result of the zod parse so the hook data is properly typed. 

Bonus points for not automatically inferring the return type and explicitly pass the zod type in the use query return type.

packages/app/features/home/TokenDetails.tsx
```diff
@@ -15,14 +15,21 @@ const TokenDetails = ({ tokenAddress }: { tokenAddress: `0x${string}` | undefine
     chainId: baseMainnet.id,
   })
 
-  // @TODO: add balance error check
   if (balance) {
+    if (balance.isError) {
+      throw new Error(`Error occurred while fetching balance. ${balance.error.message}`)
```
0xBigBoss, line 20: You're misunderstanding. If we throw an error here the UI would be unresponsive and it would not be handled since we don't have any error boundaries setup. https://www.npmjs.com/package/react-error-boundary

I meant , if an error occurred, show it as a paragraph.

```suggestion
return	(
	<Paragraph fontSize={'$9'} fontWeight={'500'} color={'$error'}>
		Error occurred while fetching balance. {balance.error.message}
</Paragraph>
)
```

packages/app/utils/useTokenPrices.ts
```diff
@@ -0,0 +1,29 @@
...
+  usd: z.number(),
+})
+
+export const TokenPricesSchema = z.object({
+  ethereum: TokenPriceSchema,
+  'send-token': TokenPriceSchema,
+  'usd-coin': TokenPriceSchema,
+})
+
+export const useTokenPrices = (): UseQueryResult<typeof TokenPricesSchema._type, Error> => {
```
0xBigBoss, line 14:
```suggestion
export const useTokenPrices = (): UseQueryResult<z.infer<typeof TokenPricesSchema>, Error> => {
```

packages/app/features/home/TokenDetails.tsx
0xLe0x: 😅

0xBigBoss (approved):
beautiful

PR #213, opened 2024-03-18
Screens/Layout Cleanup

OxLeOx:


packages/app/components/HomeTopNav.tsx
```diff
@@ -1,18 +1,28 @@
...
+import { IconArrowLeft, IconGear, IconHamburger, IconQr, IconSendLogo } from 'app/components/icons'
 import { usePathname } from 'app/utils/usePathname'
 import { useRouter } from 'solito/router'
+import { SettingsBottomSheet } from './settings/SettingsBottomSheet'
 
-export function HomeTopNav({ header, subheader }: { header: string; subheader?: string }) {
+export function HomeTopNav({
+  header,
+  subheader,
+  submenuHeader,
```
youngkidwarrior, line 12: Why have submenuHeader? Look. like it serves the same role as header

packages/app/components/HomeTopNav.tsx
```diff
@@ -26,65 +36,73 @@ export function HomeTopNav({ header, subheader }: { header: string; subheader?:
...
+            ) : media.gtLg || (media.lg && submenuHeader === '') ? (
+              <H2 fontWeight={'300'} color={'$color05'} lineHeight={32}>
+                {header}
+              </H2>
+            ) : (
+              <H2 fontSize={'$8'} fontWeight={'300'} color={'$color05'} lineHeight={'$5'}>
+                {submenuHeader}
+              </H2>
+            )}
+            {path.includes('/account') ? (
```
youngkidwarrior, line 76: Seem like a lot of this logic is specifically for account. We should move it into its own `AccountTopNav.tsx` component to avoid spaghetti

packages/app/components/HomeTopNav.tsx
```diff
@@ -26,65 +36,73 @@ export function HomeTopNav({ header, subheader }: { header: string; subheader?:
...
+                </Stack>
+              </Link>
+            )}
           </XStack>
-        ) : (
-          <H2 fontWeight={'300'} color={'$color05'} lineHeight={32}>
-            {header}
-          </H2>
+        </Container>
+        {subheader && (
```
youngkidwarrior, line 89: Subheader should be left aligned under header

packages/app/components/HomeTopNav.tsx
OxLeOx: It's for mobile header, sometimes header text is different in desktop and mobile versions.
you can take a look at the design

0xBigBoss:
going to finish this up here https://github.com/0xsend/sendapp/pull/215

PR #218, opened 2024-03-19
Fix phone number formatting on accounts page

beezybarg:
Fix phone number formatting on accounts page .


0xBigBoss (approved):

beezybarg:
Ended up finding a library to try and handle formatting  considering we are not storing the country code this will give us best practice formatting according to E.164 formats. This lib is a fork of google androids libphonenumber and is actively maintained (https://gitlab.com/catamphetamine/libphonenumber-js) . I have US numbers formatted how you would expect US numbers to look IE 

tel: 14193417547 => (419) 341-7547
tel: 19546529089 => (954) 652-9089

and also tested various other countries and they result in the correct format 

tel: 442071234567 => +44 20 7123 4567
tel: 8615555151447 => +86 155 5515 1447
tel: 380947100983 => +380 94 710 0983
tel: 5511953259340 => +55 11 95325 9340
tel: 528124129087 => +52 81 2412 9087
tel: 61489921018 => +61 489 921 018
tel: 525512545678 => +52 55 1254 5678

The numbers generated in the de-identified db are not valid numbers or country codes so I cant parse them anyway, should hold up well in prod and if we dont have a valid phone number and the user is logged in I think we have bigger issues. 

At some point will put a to-do to add country code as that would help in correctly formatting things in various ways you may want.

packages/app/utils/formatPhoneNumber.ts
```diff
@@ -1,11 +1,27 @@
...
+    number = `+${number}`
+  }
 
-  if (number.length < 4) return number
-  if (number.length < 7) return number.replace(/(\d{3})(\d{1})/, '$1-$2')
-  if (number.length < 11) return number.replace(/(\d{3})(\d{3})(\d{1})/, '$1-$2-$3')
-  return number.replace(/(\d{3})(\d{4})(\d{4})/, '$1-$2-$3')
+  try {
+    const nparse = parsePhoneNumber(number)
+    if (nparse) {
```
youngkidwarrior, line 14: NIT: could use an early return here to improve readability and avoid nesting

```
if (!nparse) {
 return
 }
 if (number.startsWith('+1') && number.length === 12) {
        return nparse.formatNational()
      }
 return nparse.formatInternational()

0xBigBoss (approved):
the libphonenumber package is new to me. this is a great start. eventually we will want to expand this formatPhoneNumber or create a similar hook for an input mask for our phone number fields.

Please commit the yarn.lock file before merging.

PR #222, opened 2024-03-20
Add logout link

0xBigBoss:


packages/app/features/account/settings/SettingsNavLink.tsx
```diff
@@ -10,22 +10,26 @@ export function SettingsNavLink({
   const location = usePathname()
   const sheet = useSheet()
   const href = props.href.toString().split('?')[0]
```
youngkidwarrior, line 12: Explain split on `?` here. None of the hrefs use query props. Is this a future proof?

youngkidwarrior (approved):
Approved but added comment for clarification

packages/app/features/account/settings/SettingsNavLink.tsx
0xBigBoss: yeah exactly.

PR #231, opened 2024-03-22
refactor: consolidate TopNav across pages, minor UI tweaks

0xBigBoss:
- Consolidate the `HomeTopNav` and `AccountTopNav` components into a single reusable `TopNav` component. The new `TopNav` takes props to customize what buttons/elements are shown.
- Update the pages and screens to use the new consolidated `TopNav` component instead of the page-specific top navs. Pass in the desired props to configure the header, subheader, and buttons.
- Make some minor styling tweaks and adjustments to margins/padding in a few places, especially on larger screen sizes.
- Remove an unused /account/profile page.
- Add a new "jet" color to the theme.

So in summary, it's primarily a refactor to consolidate redundant top navigation code into a single reusable component, along with a handful of small UI style tweaks and cleanup. The new `TopNav` component allows the top nav to be more easily customized per page while sharing the core layout and functionality.


youngkidwarrior:
Nice I'm glad to see this change. Will review right now

packages/app/components/TopNav.tsx
```diff
@@ -11,34 +12,53 @@ import {
...
+  showLogo?: boolean
+  showQrButton?: boolean
+  showSettingsButton?: boolean
+}
+
+export function TopNav({
+  header,
+  subheader,
+  showLogo = false,
+  showQrButton = false,
```
youngkidwarrior, line 32: Not the biggest fan of this implementation, since these will never bot be true

Instead we can have a component prop or a string enum like
```
  button: "settings" | "qr"
```

youngkidwarrior (approved):
Just a suggestion for a cleaner button implementation

packages/app/components/TopNav.tsx
0xBigBoss: great suggestion

PR #238, opened 2024-03-25
docs update

0xBigBoss:


youngkidwarrior (approved):
I remember you told me that not everything needs to be rerunning all the time, tests and whatnot

In the future we may want a section that explains best practices to run on a low end machine

0xBigBoss:
> I remember you told me that not everything needs to be rerunning all the time, tests and whatnot
> 
> In the future we may want a section that explains best practices to run on a low end machine

yeah i used this PR to acquaint myself again with the contributing docs. I think I'll add more pages or an effective tilt usage section in a later one.

PR #266, opened 2024-04-03
wagmi: import rainbowkit and walletconect only on web

0xBigBoss:
Adds native rainbowkit and walletconnect providers paving the path for future integration on native in a safe way (no web code)
Cleans up the default wagmi config in `@my/wagmi` removing web specific stuff
Other type and smol fixes
Drops some old wagmi config code


beezybarg (approved):
lgtm

PR #270, opened 2024-04-04
app: fix TopNav spacing, fix flaky activity test

0xBigBoss:


youngkidwarrior (approved):
Nice, assuming we will be re-add `next:web` as a dependency to `webauthn-credentials` when send accounts are done?

0xBigBoss:
> Nice, assuming we will be re-add `next:web` as a dependency to `webauthn-credentials` when send accounts are done?

nah i'm just cookin'. tryin to see if we can make it even faster since it rebuilds it a total of 4 times i think 😅

PR #274, opened 2024-04-05
Profiles Send ID 

beezybarg:
This PR Adds the send_id to the users profile, the send id represents the order in which a user signed up for an account. 
- Added new supabase tests and updated existing tests that utilize tag_search
- Added result set for tag_search to include tag_matches, phone_matches, send_id_matches 
- Phone numbers are only matched if at least 6 digits are entered.
- tag_search function now takes limit and offset param args so we can have pagination and limit results, currently limit is set to 5 until pagination is implemented. tag_search(query, limit, offset)

Frontend needs to be updated to handle the new resultset.


supabase/migrations/20240116215010_create_tag_search_function.sql
0xBigBoss, line 1: careful with this approach of modifying already applied migrations, pretty sure staging won't get properly updated since supabase thinks it's already applied.

This would work for prod though since it hasn't been applied. 

On the safe side, let's apply this in a new migration.

supabase/database.types.ts
0xBigBoss, line 1: send_id can never be null right?

packages/app/utils/getLink.ts
```diff
@@ -1,18 +1,19 @@
...
+export const getShareableLink = (linkType: string, Code: string) => {
   let baseUrl = getBaseUrl()
 
   if (!baseUrl && typeof window !== 'undefined') {
     const { protocol, hostname, port } = window.location
     baseUrl = `${protocol}//${hostname}${port ? `:${port}` : ''}`
   }
 
-  return `${baseUrl}/?referral=${referralCode}`
+  return `${baseUrl}/${linkType}/${Code}`
```
0xBigBoss, line 11: doesn't this break the referral tracking code? we don't have any test coverage for that, but I'm sure it does.

https://github.com/0xsend/sendapp/blob/dev/apps/next/pages/index.tsx#L25-L34

supabase/migrations/20240329185908_add_profiles_send_id.sql
```diff
@@ -0,0 +1,26 @@
...
+update public.profiles
+  set send_id = subquery.send_id
+  from subquery where profiles.id = subquery.user_id
+  AND profiles.send_id IS NULL;
+
+-- Create sequence for next send_ids to auto generate on insert into profiles.
+
+CREATE SEQUENCE profiles_send_id_seq;
+SELECT setval('profiles_send_id_seq', coalesce(max(send_id), 0) + 1, false) FROM profiles;
+ALTER TABLE profiles ALTER COLUMN send_id SET DEFAULT nextval('profiles_send_id_seq');
```
0xBigBoss, line 25: also make it not null. let's add an after update trigger to ensure a user can't change their send ID since users are allowed to update their profiles according to our current RLS policies.

something like this https://github.com/0xsend/sendapp/blob/dev/supabase/migrations/20230819194129_add_tags_table_rls.sql#L98-L104

0xBigBoss (commented):
didn't finish all of it. let's try to keep our feature PRs separated e.g. send IDs in one PR, and profile search stacked on top of it.

for sure we need to fix the referral code thing, unless you tested that it works?

supabase/database.types.ts
beezybarg: Corret

PR #275, opened 2024-04-05
Adding SendID to account screen, backfilling existing send ids

beezybarg:
Add SendID and backfill based on account created.

### DIFF
--- a/supabase/database-generated.types.ts
+++ b/supabase/database-generated.types.ts
@@ -237,6 +237,7 @@ export type Database = {
           is_public: boolean | null
           name: string | null
           referral_code: string | null
+          send_id: number | null
         }
         Insert: {
           about?: string | null
@@ -245,6 +246,7 @@ export type Database = {
           is_public?: boolean | null
           name?: string | null
           referral_code?: string | null
+          send_id?: number | null
         }
         Update: {
           about?: string | null
@@ -253,6 +255,7 @@ export type Database = {
           is_public?: boolean | null
           name?: string | null
           referral_code?: string | null
+          send_id?: number | null
         }
         Relationships: [
           {
--- a/supabase/database-generated.types.ts
+++ b/supabase/database-generated.types.ts
@@ -237,7 +237,7 @@ export type Database = {
           is_public: boolean | null
           name: string | null
           referral_code: string | null
-          send_id: number | null
+          send_id: number
         }
         Insert: {
           about?: string | null
@@ -246,7 +246,7 @@ export type Database = {
           is_public?: boolean | null
           name?: string | null
           referral_code?: string | null
-          send_id?: number | null
+          send_id?: number
         }
         Update: {
           about?: string | null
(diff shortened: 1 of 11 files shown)


supabase/migrations/20240329185908_add_profiles_send_id.sql
0xBigBoss, line 1: we also need to create a trigger that runs on updates to ensure that a user can't update their sendid

0xBigBoss (changes requested):

supabase/database-generated.types.ts
```diff
@@ -237,6 +237,7 @@ export type Database = {
           is_public: boolean | null
           name: string | null
           referral_code: string | null
+          send_id: number | null
```
0xBigBoss, line 240: this stupid thing, why does it thikn it can be null

supabase/tests/profile_send_id_test.sql
```diff
@@ -0,0 +1,31 @@
...
+select tests.create_supabase_user('new_profile_user2');
+
+PREPARE send_id_1 AS select send_id::int from public.profiles where id = tests.get_supabase_uid('new_profile_user1');
+PREPARE send_id_2 AS select send_id::int from public.profiles where id = tests.get_supabase_uid('new_profile_user2');
+PREPARE send_id_2d AS select send_id::int - 1 from public.profiles where id = tests.get_supabase_uid('new_profile_user2');
+select results_ne('send_id_1', 'send_id_2', 'Test new insert increments send_id and they are not the same');
+select results_eq('send_id_1', 'send_id_2d', 'Test new profile insert increments send_id by 1');
+
+-- User can not change their own send_id
+SELECT throws_ok(
```
0xBigBoss, line 18: nice. ty sir. just a nit. but this test is running as root, we also want to check when authenticated as the user.

```sql
SELECT tests.authenticate_as('new_profile_user1');
```

supabase/migrations/20240329185908_add_profiles_send_id.sql
```diff
@@ -0,0 +1,44 @@
...
+update public.profiles
+  set send_id = subquery.send_id
+  from subquery where profiles.id = subquery.user_id
+  AND profiles.send_id IS NULL;
+
+-- Create sequence for next send_ids to auto generate on insert into profiles.
+
+CREATE SEQUENCE profiles_send_id_seq;
+SELECT setval('profiles_send_id_seq', coalesce(max(send_id), 0) + 1, false) FROM profiles;
+ALTER TABLE profiles ALTER COLUMN send_id SET DEFAULT nextval('profiles_send_id_seq');
```
0xBigBoss, line 25: can you include not null and see if the types get updated accordingly?

0xBigBoss (approved):

supabase/migrations/20240329185908_add_profiles_send_id.sql
beezybarg: altered to not null after backfilling, types updated

PR #277, opened 2024-04-05
Fix Submit Button spacing

youngkidwarrior:
Uses the `<Unspaced/>` component to remove the Spacer that comes with  `<AnimatePresence/>


0xBigBoss (approved):
interesting that this fixes it. doesn't cause problems when it's submitted?

PR #306, opened 2024-04-11
Activity Screen Styles [Desktop]

stormcloud266:
Adds styling/layout for the activity screen

Screenshot 2024-04-11 at 12 05 52 PM

### DIFF
--- a/.snaplet/snaplet-client.d.ts
+++ b/.snaplet/snaplet-client.d.ts
@@ -159,6 +159,7 @@ type Override = {
       about?: string;
       referral_code?: string;
       is_public?: string;
+      send_id?: string;
       users?: string;
       referrals_referrals_referred_idToprofiles?: string;
       referrals_referrals_referrer_idToprofiles?: string;
@@ -504,6 +505,7 @@ export interface Fingerprint {
     bucket?: FingerprintRelationField;
   }
   profiles?: {
+    sendId?: FingerprintNumberField;
     i?: FingerprintRelationField;
     referralsByReferredId?: FingerprintRelationField;
     referralsByReferrerId?: FingerprintRelationField;
--- a/.snaplet/snaplet-client.d.ts
+++ b/.snaplet/snaplet-client.d.ts
@@ -159,7 +159,6 @@ type Override = {
       about?: string;
       referral_code?: string;
       is_public?: string;
-      send_id?: string;
       users?: string;
       referrals_referrals_referred_idToprofiles?: string;
       referrals_referrals_referrer_idToprofiles?: string;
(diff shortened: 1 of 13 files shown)


stormcloud266:
@0xBigBoss I had to update the snapshot for the activity screen, and the generated `snaplet.d.ts` has a couple errors. Just Wanted to touch base on that.

Screenshot 2024-04-11 at 12 35 24 PM

.snaplet/snaplet-client.d.ts
0xBigBoss, line 1: Seems like ur database is on an older migration. Be sure to rebase and run `tilt down; tilt up` every now and then so you get the latest schema changes.

packages/app/features/activity/screen.tsx
```diff
@@ -180,7 +182,9 @@ function Suggestions() {
                 <Spinner size="small" color="$send1" />
               </Avatar.Fallback>
             </Avatar>
-            <Paragraph>@{user.username}</Paragraph>
+            <Paragraph color="$white" fontFamily="$mono">
```
0xBigBoss, line 185:
```suggestion
            <Paragraph $theme-dark={{color: "$white"}} fontFamily="$mono">
```
and then do the appropriate `$theme-light` for light version.

packages/app/features/activity/screen.tsx
```diff
@@ -191,29 +195,98 @@ function Suggestions() {
...
+        <Text
+          color="$white"
+          display="none"
+          minWidth={'$14'}
+          textAlign="right"
+          $gtMd={{ display: 'inline' }}
+        >
+          {activity.time}
+        </Text>
+        <Text color="$white" minWidth={'$14'} textAlign="right">
```
0xBigBoss, line 282: same thing here, this will break on light versions.

I think there's an option to also use `$color12` too in the base light and dark themes which should auto handle light and dark switching.

0xBigBoss (commented):
thanks this looks pretty good. Let's be sure to clear out the changes to the snaplet generated code. Then, let's also ensure we handle the light <-> dark color switching. @youngkidwarrior i'm sure has some good thoughts on how to handle it best.

The type errors in the snaplet file are harmless. TBH not sure what's going on with their code and why it's not finding those schemas. Rest assured, we aren't doing anything with those schemas.

packages/app/features/activity/screen.tsx
stormcloud266: > same thing here, this will break on light versions.
> 
> I think there's an option to also use `$color12` too in the base light and dark themes which should auto handle light and dark switching.

Thanks this was very helpful. I went ahead a pushed those requested changes.

0xBigBoss (approved):
nice. looks like ur getting the hang of tamagui in no time.

PR #342, opened 2024-04-22
Update Profile search backend functions.

beezybarg:
Update Profile Search function to accomodate searching for sendid, tag or phone number , result from tag_search will now be a list of objects representing matches for sendid, phone number or tag there is also two new arguments to the function that will take limit and offset parameters and pass those values to the subqueries, this can be used to implement pagination where needed. 

Phone numbers will only match when you enter at least 7 digits, for privacy reasons. 

Example searching for '4' with a limit of 5 

```
postgres=> select * from tag_search('4',5);
-[ RECORD 1 ]---+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
send_id_matches | [{"send_id": 4, "tag_name": "brandon"}, {"send_id": 4, "tag_name": "bitches"}, {"send_id": 4, "tag_name": "2"}, {"send_id": 4, "tag_name": "beezy"}, {"send_id": 4, "tag_name": "God"}]
tag_matches     | [{"send_id": 10, "tag_name": "4"}, {"send_id": 112, "tag_name": "42"}, {"send_id": 1534, "tag_name": "420"}, {"send_id": 3068, "tag_name": "42069"}, {"send_id": 4127, "tag_name": "4tune"}]
phone_matches   | []
```

Example searching for beginning of my test number:

```
postgres=> select * from tag_search('1291622');
-[ RECORD 1 ]---+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
send_id_matches | []
tag_matches     | []
phone_matches   | [{"phone": "12916221446844", "send_id": 4, "tag_name": "brandon"}, {"phone": "12916221446844", "send_id": 4, "tag_name": "bitches"}, {"phone": "12916221446844", "send_id": 4, "tag_name": "2"}, {"phone": "12916221446844", "send_id": 4, "tag_name": "beezy"}, {"phone": "12916221446844", "send_id": 4, "tag_name": "God"}]

```

Example searching for specific tag:

```
postgres=> select * from tag_search('ethen');
-[ RECORD 1 ]---+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
send_id_matches | []
tag_matches     | [{"send_id": 2, "tag_name": "ethen", "avatar_url": "https://fjswgwdweohwejbrmiil.supabase.co/storage/v1/object/public/avatars/743d79cf-f861-4619-b7be-c46f70dc8b68/1712459659653.jpeg"}]
phone_matches   | []
```

I limited to 5 results on the frontend for now until pagination is figured out and the new screens implemented

### DIFF
--- a/.snaplet/snaplet-client.d.ts
+++ b/.snaplet/snaplet-client.d.ts
@@ -98,6 +98,7 @@ type Override = {
       qualification_start?: string;
       qualification_end?: string;
       claim_end?: string;
+      snapshot_id?: string;
       hodler_min_balance?: string;
       created_at?: string;
       updated_at?: string;
@@ -522,6 +523,7 @@ export interface Fingerprint {
     qualificationStart?: FingerprintDateField;
     qualificationEnd?: FingerprintDateField;
     claimEnd?: FingerprintDateField;
+    snapshotId?: FingerprintNumberField;
     hodlerMinBalance?: FingerprintNumberField;
     createdAt?: FingerprintDateField;
     updatedAt?: FingerprintDateField;
--- a/packages/app/features/activity/screen.tsx
+++ b/packages/app/features/activity/screen.tsx
@@ -134,12 +134,14 @@ function SearchResults() {
     return null
   }
 
+  const result = results[0] // Access the first element of the results array
+
   return (
     <YStack
       testID="searchResults"
       key="searchResults"
       animation="quick"
-      gap="$4"
+      space="$4"
       mb="$4"
       enterStyle={{
         opacity: 0,
@@ -147,29 +149,39 @@ function SearchResults() {
       }}
     >
       <H4 theme={'alt2'}>Results</H4>
-      {results.length === 0 && <Text>No results for {query}... 😢</Text>}
-      {results.map((result) => (
-        <Link key={result.tag_name} href={`/profile/${result.tag_name}`}>
-          <XStack testID={`tag-search-${result.tag_name}`} ai="center" gap="$4">
-            <Avatar size="$4" br="$4" gap="$2">
-              <Avatar.Image src={result.avatar_url} />
-              <Avatar.Fallback>
-                <Avatar>
-                  <Avatar.Image
-                    src={`https://ui-avatars.com/api.jpg?name=${result.tag_name}&size=256`}
-                  />
-                  <Avatar.Fallback>
-                    <Paragraph>??</Paragraph>
-                  </Avatar.Fallback>
-                </Avatar>
-              </Avatar.Fallback>
-            </Avatar>
-            <YStack gap="$1">
-              <Text>{result.tag_name}</Text>
-            </YStack>
-          </XStack>
-        </Link>
-      ))}
+      {result.send_id_matches?.length === 0 &&
+        result.tag_matches?.length === 0 &&
+        result.phone_matches?.length === 0 && <Text>No results for {query}... 😢</Text>}
+      {['send_id_matches', 'tag_matches', 'phone_matches'].map(
+        (key) =>
+          result[key]?.length > 0 && (
+            <>
+              <H4 theme={'alt2'}>{key}</H4>
+              {result[key].map((result) => (
+                <Link key={result.tag_name} href={`/profile/${result.tag_name}`}>
+                  <XStack testID={`tag-search-${result.tag_name}`} ai="center" space="$4">
+                    <Avatar size="$4" br="$4" space="$2">
+                      <Avatar.Image src={result.avatar_url} />
+                      <Avatar.Fallback>
+                        <Avatar>
+                          <Avatar.Image
+                            src={`https://ui-avatars.com/api.jpg?name=${result.tag_name}&size=256`}
+                          />
+                          <Avatar.Fallback>
+                            <Paragraph>??</Paragraph>
+                          </Avatar.Fallback>
+                        </Avatar>
+                      </Avatar.Fallback>
+                    </Avatar>
+                    <YStack space="$1">
+                      <Text>{result.tag_name}</Text>
+                    </YStack>
+                  </XStack>
+                </Link>
+              ))}
+            </>
+          )
+      )}
     </YStack>
   )
 }
--- a/packages/app/provider/tag-search/TagSearchProvider.tsx
+++ b/packages/app/provider/tag-search/TagSearchProvider.tsx
@@ -30,7 +30,7 @@ async function searchTags({ supabase, query }: SearchTagsArgs) {
   }
   abortController = new AbortController()
   const { data, error } = await supabase
-    .rpc('tag_search', { query })
+    .rpc('tag_search', { query, limit_val: 10 })
     .abortSignal(abortController.signal)
   return { data, error }
 }
@@ -68,8 +68,8 @@ export const TagSearchProvider = ({ children }: { children: React.ReactNode }) =
   const query = form.watch('query', '')
 
   useEffect(() => {
-    if (query.length >= 2) {
-      onSearch({ query })
+    if (query.length >= 1) {
+      onSearch({ query, limit_val: 10 })
     } else {
       setIsLoading(false)
       setError(null)
--- a/supabase/database-generated.types.ts
+++ b/supabase/database-generated.types.ts
@@ -186,6 +186,7 @@ export type Database = {
           qualification_end: string
           qualification_start: string
           snapshot_block_num: number | null
+          snapshot_id: number | null
           updated_at: string
         }
         Insert: {
@@ -204,6 +205,7 @@ export type Database = {
           qualification_end: string
           qualification_start: string
           snapshot_block_num?: number | null
+          snapshot_id?: number | null
           updated_at?: string
         }
         Update: {
@@ -222,6 +224,7 @@ export type Database = {
           qualification_end?: string
           qualification_start?: string
           snapshot_block_num?: number | null
+          snapshot_id?: number | null
           updated_at?: string
         }
         Relationships: []
@@ -916,15 +919,19 @@ export type Database = {
           address: string
           chain_id: number
           is_public: boolean
+          send_id: number
         }[]
       }
       tag_search: {
         Args: {
           query: string
+          limit_val?: number
+          offset_val?: number
         }
         Returns: {
-          avatar_url: string
-          tag_name: string
+          send_id_matches: Json
+          tag_matches: Json
+          phone_matches: Json
         }[]
       }
       update_distribution_shares: {
--- a/supabase/database.types.ts
+++ b/supabase/database.types.ts
@@ -83,7 +83,7 @@ export type Database = MergeDeep<
             address: `0x${string}`
             chain_id: number
             is_public: boolean | null
-            send_id: number | null
+            send_id: number
           }[]
           tag_search: {
             Returns: {
--- a/supabase/database.types.ts
+++ b/supabase/database.types.ts
@@ -83,8 +83,34 @@ export type Database = MergeDeep<
             address: `0x${string}`
             chain_id: number
             is_public: boolean | null
-            send_id: number
+            send_id: number | null
           }[]
+          tag_search: {
+            Returns: {
+              send_id_matches: [
+                {
+                  send_id: number | null
+                  tag_name: string
+                  avatar_url: string | null
+                },
+              ]
+              tag_matches: [
+                {
+                  tag_name: string
+                  avatar_url: string | null
+                  send_id: number | null
+                },
+              ]
+              phone_matches: [
+                {
+                  phone: string
+                  avatar_url: string | null
+                  send_id: number | null
+                  tag_name: string | null
+                },
+              ]
+            }
+          }
         }
       }
     }
--- a/supabase/migrations/20240122020348_create_profile_lookup_function.sql
+++ b/supabase/migrations/20240122020348_create_profile_lookup_function.sql
@@ -8,7 +8,8 @@ create or replace function public.profile_lookup(tag text)
     tag_name citext,
     address citext,
     chain_id integer,
-    is_public boolean)
+    is_public boolean,
+    send_id integer)
   language plpgsql
   immutable
   security definer
@@ -28,7 +29,8 @@ begin
     sa.chain_id as chain_id,
     case when current_setting('role')::text = 'service_role' then
       p.is_public
-    end as is_public
+    end as is_public,
+    p.send_id as send_id
   from
     profiles p
     join tags t on t.user_id = p.id
--- a/supabase/migrations/20240415200034_update_tag_search_function.sql
+++ b/supabase/migrations/20240415200034_update_tag_search_function.sql
@@ -1,0 +1,103 @@
+
+drop function if exists public.tag_search(q text);
+create or replace function public.tag_search(
+        query text,
+        limit_val integer = NULL,
+        offset_val integer = NULL
+    ) returns table(
+        send_id_matches jsonb,
+        tag_matches jsonb,
+        phone_matches jsonb
+    ) language plpgsql immutable security definer as $function$ begin return query --
+  select
+    COALESCE(
+        (
+            select jsonb_agg(
+                    jsonb_strip_nulls(
+                        jsonb_build_object(
+                            'avatar_url',
+                            avatar_url,
+                            'tag_name',
+                            name,
+                            'send_id',
+                            send_id
+                        )
+                    )
+                )
+            from (
+                    select p.avatar_url,
+                        t.name,
+                        p.send_id
+                    from profiles p
+                        join tags t on t.user_id = p.id
+                    where query SIMILAR TO '\d+'
+                        and p.send_id::varchar ilike '%' || query || '%'
+                    order by p.send_id
+                    LIMIT limit_val OFFSET offset_val
+                ) sub
+        ),
+        '[]'
+    ) as send_id_matches,
+    COALESCE(
+        (
+            select jsonb_agg(
+                    jsonb_strip_nulls(
+                        jsonb_build_object(
+                            'avatar_url',
+                            avatar_url,
+                            'tag_name',
+                            name,
+                            'send_id',
+                            send_id
+                        )
+                    )
+                )
+            from (
+                    select p.avatar_url,
+                        t.name,
+                        p.send_id
+                    from profiles p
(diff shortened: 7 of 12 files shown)


supabase/migrations/20240415200034_update_tag_search_function.sql
0xBigBoss, line 1: This is a great first attempt. Let's build on it now and introduce a custom composite type. This way we get types on the frontend.

```sql
drop type if exists public.tag_search_result cascade;
create type public.tag_search_result as (
                                            avatar_url text,
                                            tag_name text,
                                            send_id integer,
                                            phone text
                                        );

drop function if exists public.tag_search(query text, limit_val integer, offset_val integer);
create or replace function public.tag_search(
    query text,
    limit_val integer = NULL,
    offset_val integer = NULL
) returns table(
                   send_id_matches public.tag_search_result[],
                   tag_matches public.tag_search_result[],
                   phone_matches public.tag_search_result[]
               ) language plpgsql immutable security definer as $function$ begin return query
    select
        (
            select array_agg(row(sub.avatar_url, sub.tag_name, sub.send_id, sub.phone)::public.tag_search_result)
            from (
                     select p.avatar_url, t.name as tag_name, p.send_id, null::text as phone
                     from profiles p
                              join tags t on t.user_id = p.id
                     where query SIMILAR TO '\d+'
                       and p.send_id::varchar ilike '%' || query || '%'
                     order by p.send_id
                     limit limit_val offset offset_val
                 ) sub
        ) as send_id_matches,
        (
            select array_agg(row(sub.avatar_url, sub.tag_name, sub.send_id, sub.phone)::public.tag_search_result)
            from (
                     select p.avatar_url, t.name as tag_name, p.send_id, null::text as phone
                     from profiles p
                              join tags t on t.user_id = p.id
                     where t.name ILIKE '%' || query || '%'
                     order by (t.name <-> query)
                     limit limit_val offset offset_val
                 ) sub
        ) as tag_matches,
        (
            select array_agg(row(sub.avatar_url, sub.tag_name, sub.send_id, sub.phone)::public.tag_search_result)
            from (
                     select p.avatar_url, t.name as tag_name, p.send_id, u.phone
                     from profiles p
                              join tags t on t.user_id = p.id
                              join auth.users u on u.id = p.id
                     where query ~ '^\d{6,}$'
                       and u.phone like query || '%'
                     order by u.phone
                     limit limit_val offset offset_val
                 ) sub
        ) as phone_matches;
end;
$function$;

select * from public.tag_search('test');
```

supabase/migrations/20240415200034_update_tag_search_function.sql
```diff
@@ -0,0 +1,103 @@
...
+                        )
+                    )
+                )
+            from (
+                    select p.avatar_url,
+                        t.name,
+                        p.send_id
+                    from profiles p
+                        join tags t on t.user_id = p.id
+                    where t.name ILIKE '%' || query || '%'
```
0xBigBoss, line 61: did you mean to drop the trigram distance operator?

```sql
(t.name <<-> :query < 0.9 or t.name ilike '%' || :query || '%')
```

supabase/migrations/20240122020348_create_profile_lookup_function.sql
0xBigBoss, line 1: sadly we need to create a new migration, this won't get applied.

.snaplet/snaplet-client.d.ts
0xBigBoss, line 1: this looks like it's from the old schema and needs to be reverted.

0xBigBoss (commented):
let's add the composite type. I included an example query that does this.

also, don't forget to update the playwright specs here.  https://github.com/0xsend/sendapp/blob/f8102830061e61b9188d0eec661da7fd46c9badc/packages/playwright/tests/activity.onboarded.spec.ts

supabase/migrations/20240415200034_update_tag_search_function.sql
```diff
@@ -0,0 +1,103 @@
+
+drop function if exists public.tag_search(q text);
+create or replace function public.tag_search(
+        query text,
+        limit_val integer = NULL,
```
0xBigBoss, line 5: should we allow NULL? I'm just thinking how easy this function could tip over our database.

we shuold not allow more than 100 limit methinks.

you can validate the inputs and raise an exception like so

```sql
IF NEW.user_id <> auth.uid() THEN RAISE EXCEPTION 'Users can only create or modify tags for themselves';
```

supabase/migrations/20240415200034_update_tag_search_function.sql
beezybarg: I moved this to the order by :

order by (t.name <->query) asc

When used in an ORDER BY clause like ORDER BY (t.name <-> query) ASC, it sorts the results in ascending order of this distance, meaning the most similar strings to query will come first.

supabase/migrations/20240415200034_update_tag_search_function.sql
beezybarg: Not sure what you mean by that, this should really live in code in the backend and not on the database.

0xBigBoss: supabase can generate typescript from the composite types. not the jsonb.

Screenshot 2024-04-23 at 08 25 35

supabase/migrations/20240415200034_update_tag_search_function.sql
0xBigBoss: I added it as an OR so that similar sendtags were also returned, not just exact matching.

packages/app/features/activity/screen.tsx
```diff
@@ -134,42 +134,54 @@ function SearchResults() {
     return null
   }
 
+  const result = results[0] // Access the first element of the results array
```
0xBigBoss, line 137: should do this in the tag search provider hook.

supabase/migrations/20240415200034_update_tag_search_function.sql
```diff
@@ -0,0 +1,114 @@
...
+        offset_val integer = NULL
+    ) returns table(
+        send_id_matches jsonb,
+        tag_matches jsonb,
+        phone_matches jsonb
+    ) language plpgsql immutable security definer as $function$
+begin
+    if limit_val is not null
```
0xBigBoss, line 13:
```suggestion
    if limit_val is null or
```

supabase/migrations/20240415200034_update_tag_search_function.sql
```diff
@@ -0,0 +1,114 @@
...
+    and (
+        limit_val <= 0
+        or limit_val > 100
+    ) then raise exception 'limit_val must be between 1 and 100';
+    end if;
+    if offset_val is not null
```
0xBigBoss, line 19:
```suggestion
    if offset_val is null or
```

supabase/database-generated.types.ts
```diff
@@ -916,15 +919,19 @@ export type Database = {
...
       tag_search: {
         Args: {
           query: string
+          limit_val?: number
+          offset_val?: number
         }
         Returns: {
-          avatar_url: string
-          tag_name: string
+          send_id_matches: Database["public"]["CompositeTypes"]["tag_search_result"][]
```
0xBigBoss, line 932: beautiful

supabase/migrations/20240415200034_update_tag_search_function.sql
```diff
@@ -0,0 +1,70 @@
...
+                   phone_matches public.tag_search_result[]
+               ) language plpgsql immutable security definer as $function$
+                begin
+                    if limit_val is not null
+                    and (
+                        limit_val <= 0
+                        or limit_val > 100
+                    ) then raise exception 'limit_val must be between 1 and 100';
+                    end if;
+                    if offset_val is not null
```
0xBigBoss, line 26: same here, let's not allow null

supabase/migrations/20240415200034_update_tag_search_function.sql
```diff
@@ -0,0 +1,70 @@
...
+    select
+        (
+            select array_agg(row(sub.avatar_url, sub.tag_name, sub.send_id, sub.phone)::public.tag_search_result)
+            from (
+                     select p.avatar_url, t.name as tag_name, p.send_id, null::text as phone
+                     from profiles p
+                              join tags t on t.user_id = p.id
+                     where query SIMILAR TO '\d+'
+                       and p.send_id::varchar ilike '%' || query || '%'
+                     order by p.send_id
```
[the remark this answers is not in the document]
0xBigBoss: but i guess since we are using limit and offset we shuold incldue it to be thorough.

supabase/migrations/20240415200034_update_tag_search_function.sql
```diff
@@ -0,0 +1,70 @@
...
+                     limit limit_val offset offset_val
+                 ) sub
+        ) as send_id_matches,
+        (
+            select array_agg(row(sub.avatar_url, sub.tag_name, sub.send_id, sub.phone)::public.tag_search_result)
+            from (
+                     select p.avatar_url, t.name as tag_name, p.send_id, null::text as phone
+                     from profiles p
+                              join tags t on t.user_id = p.id
+                     where t.name ILIKE '%' || query || '%'
```
0xBigBoss, line 49: We need to keep the distance condition so that we can help users who don't know how to spell. I think I actually I had the distance constraint too strict on my first implementation.

```sql
create table tests(
    name text
);
CREATE INDEX name_idx ON tests USING GIN (name gin_trgm_ops);
insert into tests (name) values ('johndoe'), ('janedoe'), ('johnsmith'), ('jane'), ('johnwick'), ('jonwick');
select name, name <<-> 'jone' as distance from tests where name <<-> 'jone' > 0.8 or name ilike '%' || 'jone' || '%' order by (name <-> 'jone') limit 10;
select name, name <<->'john' as distance from tests where name <<->'john' > 0.8 or name ilike '%' ||'john' || '%' order by (name <->'john') limit 10;
```
```shell
CREATE TABLE
CREATE INDEX
INSERT 0 6
   name    | distance
-----------+----------
 johnsmith |      0.8
 janedoe   |    0.875
(2 rows)

   name    | distance
-----------+----------
 johndoe   |      0.5
 johnwick  | 0.555556
 johnsmith |      0.6
 jane      |      0.8
 janedoe   |    0.875
(5 rows)
```

supabase/migrations/20240415200034_update_tag_search_function.sql
```diff
@@ -0,0 +1,70 @@
...
+        ) as tag_matches,
+        (
+            select array_agg(row(sub.avatar_url, sub.tag_name, sub.send_id, sub.phone)::public.tag_search_result)
+            from (
+                     select p.avatar_url, t.name as tag_name, p.send_id, u.phone
+                     from profiles p
+                              join tags t on t.user_id = p.id
+                              join auth.users u on u.id = p.id
+                     where query ~ '^\d{6,}$'
+                       and u.phone like query || '%'
```
0xBigBoss, line 62: should we help users who forget to add the country code?

```suggestion
                       and u.phone like '%' || query || '%'
```

supabase/migrations/20240423083646_update_profile_lookup_function.sql
```diff
@@ -0,0 +1,47 @@
...
+  from
+    profiles p
+    join tags t on t.user_id = p.id
+    join send_accounts sa on sa.user_id = p.id
+  where
+    t.status = 'confirmed'::tag_status
+    and t.name = tag
+    and(p.is_public -- allow public profiles to be returned
+      or auth.uid() is not null -- allow profiles to be returned if the user is authenticated
+      or current_setting('role')::text = 'service_role') -- allow public profiles to be returned to service role
```
0xBigBoss, line 44: same here

supabase/migrations/20240423083646_update_profile_lookup_function.sql
```diff
@@ -0,0 +1,47 @@
+drop function if exists public.profile_lookup(tag text);
+create or replace function public.profile_lookup(tag text)
```
0xBigBoss, line 2: not every profile will have a sendtag.

so we can make two new URLs and two new lookup functions.


`/sendtag/{tag}` will use new tag_lookup
`/profile/{send_id}` will use new send_id_lookup

we will most likely want to create a type for this result as well and share it with both functions. maybe we can re-use the query logic easily somehow.

then we drop this old `profile_lookup`. WDYT?

supabase/migrations/20240415200034_update_tag_search_function.sql
beezybarg: Not sure on this, on one hand it would help find things better, on the other hand we dont want people phishing for numbers and tying to profiles.

packages/app/provider/tag-search/TagSearchProvider.tsx
```diff
@@ -47,16 +54,17 @@ export const TagSearchProvider = ({ children }: { children: React.ReactNode }) =
       try {
         setIsLoading(true)
         setError(null)
-        const result = await searchTags({ supabase, query })
+        const result = await searchTags({ supabase, query, limit_val: 10, offset_val: 0 })
         const { data, error } = result
+        const results = data && data.length > 0 ? data[0] : []
```
0xBigBoss, line 59: i would check the error first, but just a nit.

packages/playwright/tests/activity.onboarded.spec.ts
```diff
@@ -31,7 +31,18 @@ test('can search on activity page', async ({ page, context }) => {
   await context.route(`${SUPABASE_URL}/rest/v1/rpc/tag_search*`, async (route) => {
     expect(route.request().postDataJSON().query).toBe('test')
     await route.fulfill({
-      body: JSON.stringify(testTags.map((t) => ({ avatar_url: null, tag_name: t }))),
+      body: JSON.stringify([
```
0xBigBoss, line 34: <3

supabase/migrations/20240415200034_update_tag_search_function.sql
0xBigBoss: yeah i think ur right, prob a better solution would be, like u said, have a country code column and add an extra column to store the subscriber number which is what I thikn most people would search for.

let's tackle later.

Screenshot 2024-04-25 at 09 55 18

supabase/migrations/20240423083646_update_profile_lookup_function.sql
0xBigBoss: we can track this here https://3.basecamp.com/5743564/buckets/36023883/todos/7334062883

beezybarg: This should be resolved, new lookups to profile defined by enum id_type_enum:

`CREATE TYPE id_type_enum AS ENUM(
  'send_id',
  'tag_name',
  'referral_code',
  'address',
  'pho`

apps/next/pages/[tag].tsx
Comment (0xBigBoss) [inline apps/next/pages/[tag].tsx:1]: should we reserve our tags that could potentially match with a tag? like `/activity` or `/account`? I mean shit, someone already reserved `send` lol.

anyways. jsut thinking out loud.

apps/next/pages/[tag].tsx
```diff
@@ -70,6 +72,14 @@ export const getServerSideProps = (async (ctx: GetServerSidePropsContext) => {
       notFound: true,
     }
   }
+  if (profile?.sendid) {
```
Comment (0xBigBoss) [inline apps/next/pages/[tag].tsx:75]: why redirect? don't we want the url to be `/beezy`?

apps/next/pages/profile/[sendid].tsx
```diff
@@ -0,0 +1,85 @@
...
+    </>
+  )
+}
+
+// Profile page is not protected, but we need to look up the user profile by tag in case we have to show a 404
+export const getServerSideProps = (async (ctx: GetServerSidePropsContext) => {
+  const { sendid } = ctx.params ?? {}
+
+  // ensure identifier is valid before proceeding
+  const { success } = CheckoutTagSchema.safeParse({ name: sendid })
```
Comment (0xBigBoss) [inline apps/next/pages/profile/[sendid].tsx:30]: it's a sendid. why safe parse it? should just be a number regex.

packages/app/features/account/rewards/__snapshots__/screen.test.tsx.snap
0xBigBoss, line 1: ty, this test isn't mocking the date, so it's gonna fail everyday lol

packages/playwright/tests/profile.logged-in.spec.ts
```diff
@@ -24,17 +24,17 @@ test('logged in user needs onboarding before visiting profile', async ({ page, s
...
   assert(!!profile.about, 'profile about not found')
-  await page.goto(`/profile/${tag.name}`)
+  await page.goto(`${tag.name}`)
   expect(await page.title()).toBe('Send | Onboarding')
   await new OnboardingPage(page).completeOnboarding(expect)
 
   // @todo check that user is redirected back to profile page
-
-  await page.goto(`/profile/${tag.name}`)
+  console.log('profile', profile)
```
0xBigBoss, line 32: drop the log pls

supabase/database.types.ts
```diff
@@ -74,16 +74,25 @@ export type Database = MergeDeep<
...
         profile_lookup: {
           Returns: {
-            id: string | null
-            avatar_url: string | null
-            name: string | null
-            about: string | null
-            referral_code: string
-            tag_name: string
-            address: `0x${string}`
+            id: string
```
0xBigBoss, line 77: thanks for updating this, but i was overriding here to match the return more closely. e.g is_public is always null for non supa admins.

No need to updaate the `tag_search` here. It can use the default from `database-generated.types.ts`

supabase/migrations/20240423083646_update_profile_lookup_function.sql
0xBigBoss, line 1: nice

supabase/tests/profile_lookup_test.sql
0xBigBoss, line 1: ty

supabase/migrations/20240415200034_update_tag_search_function.sql
0xBigBoss, line 1: amazing

0xBigBoss (approved):
lgtm after we address the few things with the new /[tag] route and /profile/[sendid] route

eveything else is minor and nit picks..

PR #394, opened 2024-05-08
app: fix reward test

0xBigBoss:
makes rewards screen easier to test without relying on global mocks.


packages/app/utils/useTimeRemaining.ts
```diff
@@ -14,7 +14,6 @@ export function useTimeRemaining(targetDate?: Date, interval = 1000): TimeRemain
   )
 
   useEffect(() => {
-    setTimeRemaining(getTimeRemaining(targetDate))
```
0xBigBoss, line 17: not needed since it's set on mount

PR #437, opened 2024-05-23
Feat/account recovery

nicky-LV:
https://3.basecamp.com/5743564/buckets/36023883/todos/7213285554


apps/next/pages/challenge.tsx
nicky-LV, line 1: redundant file - can be removed

packages/api/src/routers/account-recovery/router.ts
```diff
@@ -0,0 +1,236 @@
...
+    .query(async ({ input }): Promise<RecoveryEligibilityResponse> => {
+      if (!input.phoneNumberInput) {
+        throw new TRPCError({
+          code: 'BAD_REQUEST',
+          message: formatErr('Phone number required'),
+        })
+      }
+
+      // Retrieve user from phone number
+      const { data: userData, error: getUserError } = await getUserByPhoneNumber(
```
0xBigBoss, line 40: we gonna have to rethink this one, most of these users that need acct recovery forgot their phone number cuz they used a burner. We need to accept a public address and look the user ID in the `chain_addresses` table.

packages/api/src/routers/account-recovery/router.ts
```diff
@@ -0,0 +1,236 @@
...
+        .select('*')
+        .eq('user_id', userData.id as string)
+
+      const recoveryOptions: RecoveryOptions[] = []
+      if (chainAddress || webauthnCredentials) {
+        if (chainAddress) {
+          recoveryOptions.push(RecoveryOptions.EOA)
+        }
+        if (webauthnCredentials) {
+          recoveryOptions.push(RecoveryOptions.WEBAUTHN)
```
0xBigBoss, line 67: pretty sure this option won't be necessary. it'll be used for sign in instead.

packages/api/src/routers/account-recovery/router.ts
```diff
@@ -0,0 +1,236 @@
...
+          recoveryOptions: recoveryOptions,
+        })
+      }
+      return RecoveryEligibilityResponseSchema.parse({
+        eligible: false,
+        recoveryOptions,
+        error: formatErr('No linked chain address / webauthn credentials'),
+      })
+    }),
+  getChallenge: publicProcedure.input(PhoneNumberSchema).mutation(async ({ input }) => {
```
0xBigBoss, line 80: should be able to return the challenge in the first route, no need for a 2nd one right?

packages/app/features/auth/account-recovery/eoa/RecoverWithEOA.tsx
```diff
@@ -0,0 +1,63 @@
...
+  onSignSuccess: (data: SignMessageData, variables: SignMessageVariables, context: unknown) => void
+  // https://wagmi.sh/react/api/hooks/useSignMessage#onerror
+  onSignError: (
+    error: SignMessageErrorType,
+    variables: SignMessageVariables,
+    context: unknown
+  ) => void
+}
+
+export default function RecoverWithEOA(props: Props) {
```
0xBigBoss, line 24: nice. very clean

packages/app/features/auth/account-recovery/eoa/RecoverWithEOA.tsx
```diff
@@ -0,0 +1,63 @@
+import { useAccount, useSignMessage } from 'wagmi'
+import type { SignMessageErrorType } from '@wagmi/core'
+import type { SignMessageData, SignMessageVariables } from 'wagmi/query'
+import { Button, ButtonText, Spinner } from '@my/ui'
+import { Provider } from 'app/provider'
```
0xBigBoss, line 5: why this necessary on this page? should be set higher up in the component hierarchy

packages/app/features/auth/account-recovery/screen.tsx
```diff
@@ -0,0 +1,197 @@
...
+  const verifyChallengeMutation = api.challenge.verifyChallenge.useMutation({ retry: false })
+  const router = useRouter()
+  const { resolvedTheme } = useThemeSetting()
+  const iconColor = resolvedTheme?.startsWith('dark') ? '$primary' : '$black'
+
+  useEffect(() => {
+    if (data?.recoveryOptions) {
+      // Retrieve challenge if there are recovery options
+      challengeMutation
+        .mutateAsync({ phoneNumberInput: props.phoneNumber })
```
0xBigBoss, line 49: let's convert to the connected address from wagmi

packages/api/src/routers/account-recovery/types.ts
0xBigBoss, line 1: i normally don't separate out the types, but that's just my pref

0xBigBoss: i usually keep em close to where they are defined

packages/app/features/auth/account-recovery/screen.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+  const onSignSuccess = useCallback(
+    (data: SignMessageData, variables: SignMessageVariables, context: unknown) => {
+      if (challenge) {
+        const { account } = variables
+        const address = account as `0x${string}`
+        verifyChallengeMutation
+          .mutateAsync({ signature: data, address })
+          .then((jwt) => {
+            // JWT is set via Set-Cookie header
```
0xBigBoss, line 77: that's cool. didn't know trpc would do that for us. we don't have to set the cookie client side?

0xBigBoss: we also have to think about what to do with this screen since it'll probably never make sense on native

packages/app/features/auth/sign-in/sign-in-form.tsx
```diff
@@ -37,12 +40,22 @@ export const SignInForm = () => {
...
     <FormProvider {...form}>
       {form.formState.isSubmitSuccessful ? (
-        <VerifyCode
-          phone={`${form.getValues().countrycode}${form.getValues().phone}`}
-          onSuccess={() => {
-            router.push('/')
-          }}
-        />
+        showRecoveryForm ? (
+          <AccountRecoveryScreen
```
0xBigBoss, line 44: i'd also show this before submitting the phone number. "forgot your phone number?"

supabase/migrations/20240425205531_create_auth_challege_table.sql
```diff
@@ -0,0 +1,91 @@
+-- Create a table for authentication challenge lookups
+create table "public"."auth_challenges" (
```
0xBigBoss, line 2: I'd drop the `auth_`. It's cleaner IMO.

supabase/migrations/20240425205531_create_auth_challege_table.sql
```diff
@@ -0,0 +1,91 @@
...
+    -- A unique identifier for each challenge - not persistent across upserts
+    "id" uuid not null default gen_random_uuid(),
+    -- The user_id of the user trying to recover their account
+    "user_id" uuid not null references auth.users (id) on delete cascade,
+    -- Ensure the challenge is a hex encoded 256-bit hash (sha256)
+    "challenge" citext not null constraint auth_challenges_challenge_check check (
```
0xBigBoss, line 8: i'd not make it case insensitve text. think a `bytea` would be better.

supabase/migrations/20240425205531_create_auth_challege_table.sql
```diff
@@ -0,0 +1,91 @@
...
+-- this table, aka our supanasrAdmin helper type. This prevents any spoofing
+-- attacks and disdables users from seeing who is recovering their account
+-- which could lead to them devising attack vectors to gain access to the accounts
+alter table "public"."auth_challenges" enable row level security;
+
+-- Create a copy of the users table only supabaseAdmin can view for phone
+-- number usage in challenges
+create view "public"."users" as select * from "auth"."users";
+revoke all on "public"."users" from anon, authenticated, public;
+grant all on "public"."users" to service_role;
```
0xBigBoss, line 32: service role doesn't have permissions to view `auth.users`?

supabase/migrations/20240425205531_create_auth_challege_table.sql
```diff
@@ -0,0 +1,91 @@
...
+
+-- Enable row level security, ensuring only the service role can interact with
+-- this table, aka our supanasrAdmin helper type. This prevents any spoofing
+-- attacks and disdables users from seeing who is recovering their account
+-- which could lead to them devising attack vectors to gain access to the accounts
+alter table "public"."auth_challenges" enable row level security;
+
+-- Create a copy of the users table only supabaseAdmin can view for phone
+-- number usage in challenges
+create view "public"."users" as select * from "auth"."users";
```
0xBigBoss, line 30: this view shouldn't be required.

supabase/migrations/20240425205531_create_auth_challege_table.sql
0xBigBoss: i'd rather give `service_role` permissions to select `auth.users`.

supabase/migrations/20240425205531_create_auth_challege_table.sql
```diff
@@ -0,0 +1,91 @@
...
+revoke all on "public"."users" from anon, authenticated, public;
+grant all on "public"."users" to service_role;
+
+
+-- Secure the table granting all permissions to the service role
+revoke all on "public"."auth_challenges" from anon, authenticated, public;
+grant all on table "public"."auth_challenges" to service_role;
```
0xBigBoss, line 37: `service_role` bypasses RLS

supabase/migrations/20240425205531_create_auth_challege_table.sql
```diff
@@ -0,0 +1,91 @@
...
+-- operation  Minus "user_id", the rest are then passed to a following update 
+-- function which will update the `created_at` and `expires_at` as well as
+-- the `challenge` field if the `user_id` supplied along with a new derived
+-- "id" field for the  challenge row in the table.
+alter table "public"."auth_challenges" add constraint "auth_challenges_user_id_idx"
+unique using index "auth_challenges_user_id_idx";
+
+-- Create a function to update the table and add a challenge
+create or replace function upsert_auth_challenges(
+    userid uuid,
```
0xBigBoss, line 60:
```suggestion
    user_id uuid,
```

I prefer to use `#variable_conflict use_column` and then set an alias in the `declare` section.

```sql
declare
            _user_id alias for $1 
            _auth_id uuid;
            _created timestamptz;
            _expires timestamptz;
            _new_challenge auth_challenges;
    begin
```

0xBigBoss: https://github.com/0xsend/sendapp/blob/41d76df1b91b03b6d6181b7c6ec5a8882c6fea7c/supabase/migrations/20240426211951_create_send_accounts_add_webauthn_credential.sql#L21

packages/app/utils/account-recovery/challenge.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  if (data.length === 0) {
+    return true
+  }
+  return false
+}
+
+export function challengeUserMessage(user_id: string, pii: string): string {
+  return Buffer.from(sha256(`${user_id}: ${generateRandomString(256)} :${pii}`)).toString('hex')
```
0xBigBoss, line 37:
```suggestion
  const buf = crypto.randomBytes(256);
  return Buffer.from(sha256(`${user_id}: ${buf.toString('hex')} :${pii}`)).toString('hex')
```

0xBigBoss: https://nodejs.org/docs/latest-v20.x/api/crypto.html#cryptorandombytessize-callback

packages/app/utils/jwt.ts
0xBigBoss, line 1: so simple. nice.

supabase/migrations/20240425205531_create_auth_challege_table.sql
```diff
@@ -0,0 +1,91 @@
...
+    ),
+    -- Define the creation time and expiration time for the challenge aka:
+    -- its validity window in which is can be signed and verified before another
+    -- challenge has to be made.
+    "created_at" timestamp with time zone not null default current_timestamp,
+    "expires_at" timestamp with time zone not null default current_timestamp + interval '15 minute'
+);
+
+-- Create a unique primary key
+create unique index "auth_challenges_pkey" on "public"."auth_challenges" using btree ("id");
```
0xBigBoss, line 20: pretty sure we can just use serial id. no real big benefit using uuid here.

supabase/migrations/20240425205531_create_auth_challege_table.sql
```diff
@@ -0,0 +1,91 @@
+-- Create a table for authentication challenge lookups
+create table "public"."auth_challenges" (
+    -- A unique identifier for each challenge - not persistent across upserts
+    "id" uuid not null default gen_random_uuid(),
```
0xBigBoss, line 4:
```suggestion
"id" serial primary key,
```

supabase/migrations/20240425205531_create_auth_challege_table.sql
0xBigBoss:
```suggestion
    "challenge" bytea not null
```

packages/api/src/routers/account-recovery/router.ts
```diff
@@ -0,0 +1,236 @@
...
+        cause: `${errorMsg}. ${userDataError?.message}`,
+      })
+    }
+
+    // Call the create_challenge function with the user_id and the
+    // hex encoded hashed challenge message
+    const { data: challengeData, error: challengeCreationError } = await supabaseAdmin
+      .rpc('upsert_auth_challenges', {
+        userid: userData.id,
+        challenge: challengeUserMessage(userData.id, phoneNumberInput),
```
0xBigBoss, line 108: if you change challenge to `bytea` we need to pass the hex like this `\x123456` instead of `0x123456`.

you can see some utils for doing that here https://github.com/0xsend/sendapp/blob/41d76df1b91b03b6d6181b7c6ec5a8882c6fea7c/packages/app/utils/hexToPgBase16.ts#L9

packages/app/utils/account-recovery/challenge.ts
0xBigBoss, line 37: it's also interesting, I didn't think to include any user id or sha stuff in the message. guess this works too. AFAIK they only recommend the challenges be randomly generated. https://www.w3.org/TR/webauthn-1/#cryptographic-challenges

packages/api/src/routers/account-recovery/router.ts
nicky-LV: commit [https://github.com/0xsend/sendapp/pull/437/commits/419771c7af5aefb62461ae27b5fd85dd59994544](url)

Account recovery is now via an `identifier` instead of a phone number. An `identifier` can be a user's EOA / webauthn pubkey

packages/api/src/routers/account-recovery/router.ts
nicky-LV: commit [link](https://github.com/0xsend/sendapp/pull/437/commits/419771c7af5aefb62461ae27b5fd85dd59994544)

combined eligibility and getChallenge api routes.

The API still includes the webauthn recovery option (the sign in w/ passkey feature will use the same API)

packages/api/src/routers/account-recovery/router.ts
nicky-LV: commit [link](https://github.com/0xsend/sendapp/pull/437/commits/419771c7af5aefb62461ae27b5fd85dd59994544)

merged eligibility / getChallenge api routes

packages/api/src/routers/account-recovery/types.ts
nicky-LV: I found issues importing from files on the client-side, when the file imports supabaseAdmin.

Example:
Say I create a zod schema for trpc requests and place it in a `router.ts` file. 

If I then want to import the type of the zod schema from the client, I import from the `router.ts` file, but the file also imports `supabaseAdmin` as a side-effect - leading to issues.

In this case, I separated the types to resolve this issue.
Good point though - `schemas.ts` isn't used in other files par the `router.ts` files - I'll combine here.

nicky-LV: commit [https://github.com/0xsend/sendapp/pull/437/commits/6ab0fc87b2e30d916dff6484534609826e986639](link)

combined schemas in `schemas.ts` into the account router `router.ts` + clean-ups

packages/app/features/auth/account-recovery/screen.tsx
nicky-LV: > that's cool. didn't know trpc would do that for us. we don't have to set the cookie client side?

yep. the browser handles the `Set-Cookie` header, with good support across the board:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#syntax

nicky-LV: > we also have to think about what to do with this screen since it'll probably never make sense on native

yep - screen removed. the account-recovery "screen" is not a standalone screen, it is a component in the auth screens now.

packages/app/features/auth/sign-in/sign-in-form.tsx
nicky-LV: > i'd also show this before submitting the phone number. "forgot your phone number?"

commit [link](419771c7af5aefb62461ae27b5fd85dd59994544)

supabase/migrations/20240425205531_create_auth_challege_table.sql
nicky-LV: commit [link] (06edbac53133bd42e148d7343fc8458eec412d37)

supabase/migrations/20240425205531_create_auth_challege_table.sql
nicky-LV: commit [link](06edbac53133bd42e148d7343fc8458eec412d37)

done.

supabase/migrations/20240425205531_create_auth_challege_table.sql
nicky-LV: > i'd rather give `service_role` permissions to select `auth.users`.

`service_role` already has `auth.users` perms, via: 
supabase/migrations/20230819194070_allow_service_role_update_auth_users.sql

`grant all on table auth.users to service_role;`

nicky-LV: Kindly let me know what you think of the updated schema: 

supabase/migrations/20240425205531_create_auth_challege_table.sql

commit [link](https://github.com/0xsend/sendapp/pull/437/commits/06edbac53133bd42e148d7343fc8458eec412d37)

supabase/migrations/20240425205531_create_auth_challege_table.sql
nicky-LV: removed

commit [link](https://github.com/0xsend/sendapp/pull/437/commits/06edbac53133bd42e148d7343fc8458eec412d37)

supabase/migrations/20240425205531_create_auth_challege_table.sql
nicky-LV: kindly check this [commit](https://github.com/0xsend/sendapp/pull/437/commits/06edbac53133bd42e148d7343fc8458eec412d37), do the roles / rls of the `challenges` look ok?

supabase/migrations/20240425205531_create_auth_challege_table.sql
nicky-LV: changes made. please review this [commit](https://github.com/0xsend/sendapp/pull/437/commits/06edbac53133bd42e148d7343fc8458eec412d37)

packages/app/utils/account-recovery/challenge.ts
nicky-LV: I lacked the randomBytes API, found a workaround:
```
export function generateChallenge(): Uint8Array {
  const buffer = new Uint8Array(64)
  crypto.getRandomValues(buffer)
  return buffer
}
```

supabase/migrations/20240425205531_create_auth_challege_table.sql
nicky-LV: changed. [commit](https://github.com/0xsend/sendapp/pull/437/commits/06edbac53133bd42e148d7343fc8458eec412d37)

supabase/migrations/20240425205531_create_auth_challege_table.sql
nicky-LV: changed. [commit](https://github.com/0xsend/sendapp/pull/437/commits/06edbac53133bd42e148d7343fc8458eec412d37)

packages/api/src/routers/account-recovery/router.ts
nicky-LV: changed to bytea. seems to provide the correct challenge format:

image

packages/app/utils/account-recovery/challenge.ts
nicky-LV: removed. challenges are now randomly generated (see related comment for commit hash)

nicky-LV: please review [commit](https://github.com/0xsend/sendapp/pull/437/commits/06edbac53133bd42e148d7343fc8458eec412d37)

packages/api/src/routers/account-recovery/router.ts
```diff
@@ -0,0 +1,244 @@
...
+        logger('verifyChallenge:invalid-user-or-challenge')
+        throw new TRPCError({
+          code: 'NOT_FOUND',
+          message: formatErr('Could not retrieve challenge'),
+          cause: getChallengeError.message,
+        })
+      }
+
+      // Validate challenge isn't expired
+      isChallengeExpired(challengeData.id, logger)
```
0xBigBoss, line 98:
```suggestion
      await isChallengeExpired(challengeData.id, logger)
```

packages/api/src/routers/account-recovery/router.ts
```diff
@@ -0,0 +1,244 @@
...
+    return challengeData as ChallengeResponse
+  }),
+  verifyChallenge: publicProcedure
+    .input(VerifyChallengeRequestSchema)
+    .mutation(async ({ input, ctx }): Promise<VerifyChallengeResponse> => {
+      const { recoveryType, identifier, signature } = input
+
+      // TODO: base64 decode signature
+
+      if (!recoveryType || !identifier || !signature) {
```
0xBigBoss, line 76: minor thing but this shouldn't be required due to the non-nullable and required fields in the zod type.

packages/app/features/auth/account-recovery/account-recovery.tsx
```diff
@@ -0,0 +1,147 @@
...
+        .then((challengeResponse) => {
+          return challengeResponse as ChallengeResponse
+        })
+        .catch((err) => {
+          setError(err)
+        })
+    }
+  }
+
+  const onSignSuccess = (
```
0xBigBoss, line 52:
```suggestion
  const onSignSuccess = async (
```

packages/app/features/auth/account-recovery/account-recovery.tsx
```diff
@@ -0,0 +1,147 @@
...
+    data: SignMessageData,
+    variables: SignMessageVariables,
+    context: unknown
+  ) => {
+    const { account } = variables
+    const address = account as string
+
+    verifyChallengeMutation
```
0xBigBoss, line 60:
```suggestion
    await verifyChallengeMutation
```

packages/app/utils/account-recovery.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const getChainAddress = async (chainAddress: string) => {
+  return await supabaseAdmin
+    .from('chain_addresses')
+    .select('*')
+    .eq('address', chainAddress)
+    .single()
+}
+
+export const getPasskey = async (publicKey: string) => {
```
0xBigBoss, line 12: we may have to rethink this challenges by user id since i don't think users will know how to find their public key and storing it client side isn't guaranteed. 

When we get to sign in with passkey, i think we present anon users with a challenge, they sign with passkey, then we pass back the signature, challenge, and the credential info (such as passkey name and credential id). 

After signing a challenge with a passkey, [the passkey name is passed back](https://github.com/0xsend/sendapp/blob/e80027698b9fb0551cdc95e8eddbe29a0e623cba/packages/daimo-expo-passkeys/src/ExpoPasskeysModule.web.ts#L103-L108). Or we can update the expo passkey API to return more credential info as well. 

Another simpler approach, would be to recover the public key from the signature by using the [p256 library here](https://github.com/paulmillr/noble-curves?tab=readme-ov-file#ecdsa-public-key-recovery--extra-entropy).

nicky-LV: Yep! Noticed similar yesterday. Using expo, we only have access to the user's pubkey **after** signing (meaning that anon users must be able to request challenges).

Currently refactoring the challenges table / API to allow anon users to request challenges.

nicky-LV: [commit](https://github.com/0xsend/sendapp/pull/437/commits/b993c2318546f6791fc1633617de5d7914a9f03f)

packages/app/utils/account-recovery.ts
```diff
@@ -17,8 +17,33 @@ export const getPasskey = async (publicKey: string) => {
...
+ *
+ *
+ * @param {int} [maxRetries=3] Maximum number of retries
+ * @returns Challenge
+ */
+export const tryInsertChallenge = async (maxRetries: int = 3) => {
+  if (maxRetries === 0) {
+    return
+  }
+  const result = await supabaseAdmin
```
0xBigBoss, line 31: chances of getting a duplicate are nearly impossible. it's a good idea to have retry mechanism. Much more likely you'd get a network error or something.

retries usually need to be spread out with a delay. Luckily there is a great library to help us with that.

```ts
import { withRetry } from 'viem'

await withRetry(async () => {
//....
}, {
 delay: 500,
retryCount: 3
})
```

packages/api/src/routers/account-recovery/router.ts
```diff
@@ -0,0 +1,221 @@
...
+
+    return challengeData as ChallengeResponse
+  }),
+  verifyChallenge: publicProcedure
+    .input(VerifyChallengeRequestSchema)
+    .mutation(async ({ input, ctx }): Promise<VerifyChallengeResponse> => {
+      const { recoveryType, identifier, signature, challengeId } = input
+
+      const { data: challengeData, error: getChallengeError } = await getChallengeById(challengeId)
+      console.log(challengeData, getChallengeError, challengeId)
```
0xBigBoss, line 62:
```suggestion
      logger(challengeData, getChallengeError, challengeId)
```

0xBigBoss:
getting this build error on your PR.


Screenshot 2024-05-31 at 23 41 53

supabase/migrations/20240425205531_create_auth_challege_table.sql
nicky-LV: removed rls rules for the `challenges` table. anon users should have no restrictions inserting / selecting challenges.

packages/app/features/auth/account-recovery/eoa/RecoverWithEOA.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export default function RecoverWithEOA(props: Props) {
+  const { signMessage } = useSignMessage()
+  const { address } = useAccount()
+
+  // https://wagmi.sh/react/api/hooks/useSignMessage#onsuccess
+  const onSuccess = async (
+    data: SignMessageData,
+    variables: SignMessageVariables,
+    context: unknown
```
0xBigBoss, line 29: can remove.

packages/app/features/auth/account-recovery/account-recovery.tsx
```diff
@@ -0,0 +1,143 @@
...
+        })
+        .catch((err) => {
+          setError(err)
+        })
+    },
+    [validateSignatureMutateAsync, router]
+  )
+
+  const onSignError = () => {
+    setError({
```
0xBigBoss, line 68: need to show the user the error.


```suggestion
    const onSignError = (e: SignMessageErrorType) => {
    setError({
      message: `Failed to sign challenge. ${e.message.split('.')[0]}`,
    })
  }
```

0xBigBoss: what it looks like for me on first try trying to recover EOA with this suggestion.

Screenshot 2024-06-03 at 08 09 23

packages/app/features/auth/account-recovery/eoa/RecoverWithEOA.tsx
```diff
@@ -0,0 +1,63 @@
...
+        }
+      )
+    }
+  }
+
+  return (
+    <Provider>
+      <OpenConnectModalWrapper>
+        <Button onPress={onPress} width={'$12'}>
+          <ButtonText>EOA</ButtonText>
```
0xBigBoss, line 58: Need to account for `isConnected` in the `useAccount` hook. For some reason, this onPress isn't bubbling up correctly for the connect modal open.

```ts
const { account, isConnected } = useAccount()
//...
<Button onPress={isConnected ? onPress : () => {}} width={'$12'}>
```

packages/app/features/auth/account-recovery/eoa/RecoverWithEOA.tsx
```diff
@@ -0,0 +1,63 @@
...
+      identifier: variables.account as string,
+      challengeId: props.challengeData.id,
+    })
+  }
+
+  const onPress = async () => {
+    if (props.challengeData.challenge) {
+      signMessage(
+        {
+          message: props.challengeData.challenge,
```
0xBigBoss, line 43: Let's clean up this challenge and convert to a 0x hex string. Also add a preamble to the challenge.
```suggestion
          message: byteaToHex(props.challengeData.challenge), 
```

0xBigBoss: Screenshot 2024-06-03 at 08 25 11

packages/api/src/routers/account-recovery/router.ts
```diff
@@ -0,0 +1,265 @@
...
+} from 'app/utils/account-recovery'
+import { mintAuthenticatedJWTToken } from 'app/utils/jwt'
+import { verifyMessage, hexToBytes } from 'viem'
+import { verifySignature } from 'app/utils/userop'
+import { COSEECDHAtoXY } from 'app/utils/passkeys'
+import { pgBase16ToHex } from 'app/utils/pgBase16ToHex'
+
+const logger = debug('api:routers:account-recovery')
+
+export const GetChallengeResponseSchema = z.object({
```
0xBigBoss, line 26: no need for the other fields.

```suggestion
export const GetChallengeResponseSchema = z.object({
  challenge: z.string(),
  expires_at: z.string(),
})
```

packages/api/src/routers/account-recovery/router.ts
```diff
@@ -0,0 +1,265 @@
...
+          message: formatErr(
+            `Unauthorized. Ensure that you are signing with the ${recoveryType} you signed up with`
+          ),
+          cause: 'Signature failed verification',
+        })
+      }
+
+      // identify user from identifier
+      const userId = await getUserIdByIdentifier(recoveryType, identifier)
+      const jwt = mintAuthenticatedJWTToken(userId)
```
0xBigBoss, line 118: let's add a `@todo` comment reminding us to add audit capabilities for the jwt's generated via this method. currently we're blind since gotrue doesn't know about these jwts.

maybe add the links I shared awhile ago on the JS sdk calls to refresh the session? 

anyways feel free to paste this context as a comment into the code, no need to hold up the PR for this.

supabase/migrations/20240425205531_create_auth_challege_table.sql
```diff
@@ -0,0 +1,36 @@
...
+    "challenge" bytea not null,
+    "created_at" timestamp with time zone not null default current_timestamp,
+    "expires_at" timestamp with time zone not null default current_timestamp + interval '15 minute',
+    unique (challenge)
+);
+
+
+alter table "public"."challenges" enable row level security;
+grant all on table "public"."challenges" to service_role;
+grant select, insert, update on "public"."challenges" to anon, authenticated;
```
0xBigBoss, line 12: don't think we want this one. we want users using our tRPC API, not through the postgrest API (via supabase-js).

supabase/migrations/20240425205531_create_auth_challege_table.sql
```diff
@@ -0,0 +1,36 @@
...
+    "id" serial primary key,
+    "challenge" bytea not null,
+    "created_at" timestamp with time zone not null default current_timestamp,
+    "expires_at" timestamp with time zone not null default current_timestamp + interval '15 minute',
+    unique (challenge)
+);
+
+
+alter table "public"."challenges" enable row level security;
+grant all on table "public"."challenges" to service_role;
```
0xBigBoss, line 11: shouldn't be required.
```suggestion
grant all on table "public"."challenges" to service_role;
```

supabase/migrations/20240425205531_create_auth_challege_table.sql
0xBigBoss: if no policies are enabled, by default anon and authenticated users can't do anything with the table.

packages/app/utils/userop.ts
```diff
@@ -146,7 +146,7 @@ export function generateChallenge({
  * struct for the SendVerifier contract.
  */
 export async function signChallenge(challenge: Hex) {
-  assert(isHex(challenge) && challenge.length === 80, 'Invalid challenge')
+  // assert(isHex(challenge) && challenge.length === 80, 'Invalid challenge')
```
0xBigBoss, line 149: can remove

supabase/migrations/20240425205531_create_auth_challege_table.sql
```diff
@@ -0,0 +1,36 @@
+create table "public"."challenges" (
+    "id" serial primary key,
+    "challenge" bytea not null,
```
0xBigBoss, line 3: sorry, i don't know why I didn't suggest this earlier, but it probably would have been much easier to just use `pg_crypto.gen_random_bytes`


```suggestion
    "challenge" bytea not null default gen_random_bytes(64),
```

https://www.postgresql.org/docs/current/pgcrypto.html#PGCRYPTO-RANDOM-DATA-FUNCS

packages/api/src/routers/account-recovery/router.ts
```diff
@@ -0,0 +1,265 @@
...
+      const encodedJwt = encodeURIComponent(JSON.stringify([jwt, null, null, null, null, null]))
+
+      logger('api:accountRecoveryRouter:signature-verified')
```
0xBigBoss, line 121: let's turn this into a `console.log` and include some context about the request. the debug logger isn't enabled on the server. 

```suggestion
console.log(`account recovered using ${recoveryType} user:${userID} 
```

this way we have minimal monitoring and a short audit trail.

supabase/migrations/20240425205531_create_auth_challege_table.sql
0xBigBoss: it's fine to leave as is too. just informational

packages/app/features/auth/account-recovery/eoa/RecoverWithEOA.tsx
0xBigBoss: is this still required? can we remove it?

supabase/migrations/20240425205531_create_auth_challege_table.sql
0xBigBoss: RLS is managed via policies. By default, these roles should already have these permissions on tables in the public schema.

The correct way would be to create a policy and define the condition for who has access. 

```sql
create policy "Individuals can view their own todos."
on todos for select
using ( (select auth.uid()) = user_id );
```

supabase/tests/rls_public_test.sql
```diff
@@ -42,7 +42,7 @@ $$ LANGUAGE sql;
 
 select tests.rls_enabled(
     'public',
-    ARRAY ['distribution_verifications_summary']
+    ARRAY ['distribution_verifications_summary', 'users']
```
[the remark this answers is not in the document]
0xBigBoss: be sure to remove

packages/app/utils/userop.ts
nicky-LV: removed.

supabase/migrations/20240425205531_create_auth_challege_table.sql
nicky-LV: refactored. thanks! much cleaner this way.

packages/app/features/auth/account-recovery/eoa/RecoverWithEOA.tsx
nicky-LV: image

PR #439, opened 2024-05-29
global search bar

stormcloud266:


stormcloud266:
* **#439** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/439?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `dev`

<h2></h2>

stormcloud266:
@youngkidwarrior yo is this sort of what you were thinking with the search bar using the url?

packages/app/utils/useSearchResultHref.tsx
```diff
@@ -0,0 +1,21 @@
...
+    case '/send':
+      return `/send?${new URLSearchParams({
+        ...sendParams,
+        recipient: item.tag_name,
+      }).toString()}`
+    default: {
+      throw new Error('Unhandled path in `useSearchResultHref`')
+    }
+  }
+}
```
youngkidwarrior, line 21: <i class='graphite__hidden'>[Re: line 22]<br/><br/></i>

Looks pretty good. Just a suggestion, possibly for a less verbose solution

I think it's handling a bit too much.  Pass the route and params down via props from the screen through the Search component. The screen should be aware of its own path, right?

So you could have a function like this maybe

`useSearchBarHref(item: SearchResultCommonType, to: Route, queryParams?: RouteParams)`

I think we could make the type system work for us here with a little work too, so that we always match the right path to the right params.

Can we maybe do some pairing on this. I'd like a nice, extendable API for this<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/439?utm_source=unchanged-line-comment">Graphite</a>.</span>

packages/app/components/SearchBar.tsx
```diff
@@ -178,18 +199,40 @@ function SearchFilterButton({
...
+    case '/send':
+      return `/send?${new URLSearchParams({
+        ...queryParams,
+        recipient: item.tag_name,
+      }).toString()}`
+    default: {
+      throw new Error('Unhandled path in `useSearchResultHref`')
+    }
+  }
+}
```
stormcloud266, line 219: @youngkidwarrior so i'm able to remove the `useSendParams` and `usePathname` from the function, it's just the manual handling of the the return href string since the two cases are much different

youngkidwarrior: yeah I like this but let me see if we can get the type system to work for us. 

So, for example, when the `to` type is `send`, then we only use the send query params.

packages/app/features/activity/screen.test.tsx
```diff
@@ -5,6 +5,10 @@ import { ActivityScreen } from './screen'
 
 jest.mock('app/features/activity/utils/useActivityFeed')
 
+jest.mock('expo-router', () => ({
+  usePathname: jest.fn(),
+}))
```
stormcloud266, line 10: If I don't mock this the test fails with this error:

Screenshot 2024-06-01 at 12 05 17 PM

It looks like ever other place `usePathname` is used is a `.web.tsx` file. Not sure if there is a better way to fix this error

0xBigBoss: yeah this is expected, i thikn there are platform specific APIs inside `expo-router` that don't work in the jest environment. [we can potentially just add this to the global setup file](https://github.com/0xsend/sendapp/blob/0c0f5c2c19ca3ccf0dfe5b9d7c0221ddeab1ea50/packages/app/jest.setup-after-env.ts#L11).

you can see an example mock here https://github.com/0xsend/sendapp/blob/0c0f5c2c19ca3ccf0dfe5b9d7c0221ddeab1ea50/packages/app/features/auth/onboarding/screen.test.tsx#L16-L26

packages/app/features/activity/screen.tsx
```diff
@@ -41,12 +33,11 @@ export function ActivityScreen() {
     <TagSearchProvider>
       <YStack f={1} width={'100%'} pb="$4">
         <View>
-          <YStack width={'100%'} gap="$size.1.5" $gtSm={{ gap: '$size.2.5' }}>
-            <H4 color="$gray11Light" fontFamily={'$mono'} fontWeight={'500'} size={'$5'}>
-              SEARCH BY
-            </H4>
-            <Search />
-          </YStack>
+          <Container>
```
0xBigBoss, line 36: is this container still required? I thought we hoisted it up to the layout?

packages/app/features/activity/screen.tsx
0xBigBoss, line 1: sooo clean. happy we got to clean this up.

packages/app/utils/useSearchResultHref.tsx
0xBigBoss, line 1: very clean. please add this test. we also need to support sending to users who do not have a sendtag (just a sendid). Possibly in the future, we allow sending to just plain addresses.


```ts
// packages/app/utils/useSearchResultHref.test.tsx
import { describe } from '@jest/globals'
import { useSearchResultHref } from './useSearchResultHref'
import { usePathname } from 'app/utils/usePathname.native'
import type { SearchResultCommonType } from 'app/components/SearchBar'
import type { SendScreenParams } from 'app/routers/params'

const sendParams: SendScreenParams = {
  recipient: 'alice',
  amount: '0.01',
  sendToken: '0x123',
}

jest.mock('app/routers/params', () => ({
  useSendScreenParams: jest.fn().mockReturnValue([sendParams]),
}))

jest.mock('expo-router', () => ({
  usePathname: jest.fn(),
}))

jest.mock('app/utils/usePathname.native', () => ({
  usePathname: jest.fn(),
}))

const item = {
  // avatar_url: string;
  avatar_url: 'https://i.pravatar.cc/500?u=alice',
  // tag_name: string;
  tag_name: 'alice',
  // send_id: number;
  send_id: 12530,
  // phone: string;
  phone: '+1234567890',
} as SearchResultCommonType

describe('useSearchResultHref', () => {
  it('should return the correct href for activity screen', () => {
    // @ts-expect-error mock
    usePathname.mockReturnValue('/activity')
    const href = useSearchResultHref(item)
    expect(href).toBe('/profile/12530')
  })
  it('should return the correct href for send screen', () => {
    // @ts-expect-error mock
    usePathname.mockReturnValue('/send')
    const href = useSearchResultHref(item)
    expect(href).toBe('/send?recipient=alice&amount=0.01&sendToken=0x123')
  })
  // @fixme
  // it('should return the correct href for send screen when user does not have sendtag', () => {
  //   // @ts-expect-error mock
  //   usePathname.mockReturnValue('/send')
  //   const href = useSearchResultHref({ ...item, tag_name: null })
  //   expect(href).toBe('/send?recipient=alice&amount=0.01&sendToken=0x123')
  // })
})
```

0xBigBoss (approved):
just a few Qs and one ask. otherwise lgtm

packages/app/utils/useSearchResultHref.tsx
0xBigBoss: Let’s also handle some of the edge cases. I’m not sure of the expected API with the send params here.

packages/app/features/activity/screen.tsx
stormcloud266: You're right. I missed that update! I was able to remove several Containers

packages/app/utils/useSearchResultHref.tsx
stormcloud266: This has been added. Updated the activity screen test as well and now they are all passing

packages/app/utils/useSearchResultHref.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const [sendParams] = useSendScreenParams()
+
+  switch (path) {
+    case '/activity':
+      return `/profile/${item.send_id}`
+    case '/send':
+      return `/send?${new URLSearchParams({
+        ...JSON.parse(JSON.stringify(sendParams)), //JSON makes sure we don't pass undefined values
+        recipient: item.tag_name || item.send_id,
+      }).toString()}`
```
stormcloud266, line 16: @youngkidwarrior Big boss asked me to handle cases where `item.tag_name` doesn't exist, and I just wanted to make sure this is what you are expecting in the params in that case

PR #468, opened 2024-06-15
feat: send checks contract + unit tests

nicky-LV:


0xBigBoss (commented):
try running `yarn workspace @my/contracts lint:fix` to fix the lint errors

packages/contracts/src/SendCheck.sol
```diff
@@ -0,0 +1,66 @@
...
+contract SendCheck {
+    using SafeERC20 for IERC20;
+
+    mapping(IERC20 => mapping(address => Check)) public checks;
+
+    event CheckCreated(Check check, IERC20 token);
+    event CheckClaimed(Check check, IERC20 token, address redeemer);
+
+    function createCheck(IERC20 token, address ephemeralAddress, uint256 amount) external {
+        require(address(token) != address(0), "Cannot create /send Check: invalid token address");
```
0xBigBoss, line 24: drop the preamble, it's fine now for debugging, but it costs more gas I believe to increase the size of the revert message.


```suggestion
        require(address(token) != address(0), "invalid token address");
```

packages/contracts/src/SendCheck.sol
```diff
@@ -0,0 +1,66 @@
...
+struct Check {
+    address ephemeralAddress;
+    address from;
+    uint256 amount;
+}
+
+contract SendCheck {
+    using SafeERC20 for IERC20;
+
+    mapping(IERC20 => mapping(address => Check)) public checks;
```
0xBigBoss, line 18: what about adding token address to the `Check` struct? and just keeping the checks storage a mapping of address to Check structs?

This simplifies the claims as well.

0xBigBoss (approved):
Since these contracts don't exist on base sepolia or base mainnet, you'll probably want to include the deploy script in tilt for the next part.

here's a starting point that deploys our send account factory, https://github.com/0xsend/sendapp/blob/543cef7a1887f6ac17632ab48022f910de5bf16a/packages/contracts/script/anvil-add-send-account-factory-fixtures.ts.

Then, how it's integrated into [the tiltfile](https://github.com/0xsend/sendapp/blob/006d62c7ff7ef2e5f996a50334e4445dbc7efcdf/Tiltfile#L385-L396). Just don't forget to include it as a `resource_deps` for our [`anvil:fixtures` resource](anvil:fixtures) else things will not startup correctly in CI.

Otherwise, LGTM! I'll let you make the call on the `Check` struct and the API for claiming https://github.com/0xsend/sendapp/pull/468/files#r1641422705.

packages/contracts/src/SendCheck.sol
nicky-LV: agreed - does simplify logic. added in this [commit](https://github.com/0xsend/sendapp/pull/468/commits/30040e882acee07cd18283473f872151916e090c)

packages/contracts/script/anvil-add-send-check-fixtures.ts
```diff
@@ -0,0 +1,14 @@
+import 'zx/globals'
+
+const RPC_URL = 'http://localhost:8546'
+void (async function main() {
+  console.log(chalk.blue('Deploying SendCheck contract...'))
+  await $`forge script ./script/DeploySendCheck.s.sol:DeploySendCheckScript \
```
0xBigBoss, line 6: Setting auto mine like I did in the other deploys would shave off a second or two. This way it waits until the block is mined. All good either way.

PR #589, opened 2024-07-14
Feat/send checks mvp

nicky-LV:


apps/next/pages/checks/claim/[payload].tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  const onError = (e: Error) => {
+    // TODO: implement onError
+    console.log(e)
+  }
+
+  return (
+    <>
+      <Head>
+        <title>Claim /send check</title>
```
Comment (0xBigBoss) [inline apps/next/pages/checks/claim/[payload].tsx:22]:
```suggestion
        <title>Send | Claim Send Check</title>
```

The / will mess with SEO

packages/app/features/checks/utils/checkUtils.ts
```diff
@@ -0,0 +1,136 @@
...
+ * @returns {string} The generated URL for claiming the check.
+ */
+export const encodeClaimCheckUrl = (
+  senderId: string,
+  ephemeralKeyPair: EphemeralKeyPair
+): string => {
+  const encodedPayload = encodeURIComponent(
+    `${senderId}:${ephemeralKeyPair.ephemeralPrivateKey}:${ephemeralKeyPair.ephemeralAddress}`
+  )
+  return `/checks/claim/${encodedPayload}`
```
0xBigBoss, line 23: Add it to the URL fragment. We don't want the browser to send it to the server since we would potentially have knowledge of the key and want to avoid. This would be a security risk for us.

```suggestion
  return `/checks/claim#${encodedPayload}`
```

apps/next/pages/checks/claim/[payload].tsx
Comment (nicky-LV) [reply apps/next/pages/checks/claim/[payload].tsx:22]: addressed in PR
https://github.com/0xsend/sendapp/pull/617

packages/app/features/checks/utils/checkUtils.ts
nicky-LV: addressed in PR
https://github.com/0xsend/sendapp/pull/617

PR #596, opened 2024-07-15
use deposit label

stormcloud266:


stormcloud266:
* **#596** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/596?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `dev`

<h2></h2>

packages/app/utils/activity.ts
```diff
@@ -103,8 +103,10 @@ export function eventNameFromActivity(activity: Activity) {
   const isTransferOrReceive =
     isSendAccountTransfersEvent(activity) || isSendAccountReceiveEvent(activity)
   switch (true) {
-    case isTransferOrReceive && !!to_user?.id:
+    case isSendAccountTransfersEvent(activity) && !!to_user?.id:
       return 'Received'
+    case isSendAccountReceiveEvent(activity) && !!to_user?.id:
+      return 'Deposit'
```
stormcloud266, line 109: Is this always a deposit?

stormcloud266:
There was a note that this is for "eth". Not sure what the difference is if they deposit something else. 

```ts
export enum Events {
/ ...
  /**
   * Send account receives ETH
   */
  SendAccountReceive = 'send_account_receives',
}
```

packages/app/utils/activity.ts
0xBigBoss: It is not. `send_account_receives` are for ETH. 

The way to know it's a `Deposit` would be either `isSendAccountTransfersEvent` or `isSendAccountReceiveEvent`.

And, `from_user?.id` would be null.

stormcloud266: Look likes, at least locally, `from_user?.id` can be null, but `from_user?.send_id` wasn't...  Ended up just checking if `from_user` itself was null, and that seemed to give the correct results.

0xBigBoss (approved):

PR #623, opened 2024-07-24
open modal button

stormcloud266:
add a little wallet button that pops the connect modal

### DIFF
--- a/packages/app/features/deposit/web3/screen.tsx
+++ b/packages/app/features/deposit/web3/screen.tsx
@@ -47,12 +47,6 @@ export function DepositWeb3Screen() {
   const { open: openConnectModal } = useWeb3Modal()
   const { isConnected, chainId, chain } = useAccount()
 
-  useEffect(() => {
-    if (!isConnected) {
-      openConnectModal?.()
-    }
-  }, [isConnected, openConnectModal])
-
   if (!isConnected) {
     return (
       <Wrapper>
@@ -70,6 +64,8 @@ export function DepositWeb3Screen() {
         >
           Connect to Deposit
         </Button>
+
+        <DepositAddressWrapper />
       </Wrapper>
     )
   }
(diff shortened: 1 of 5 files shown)


stormcloud266:
* **#623** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/623?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `dev`

<h2></h2>

youngkidwarrior:
Honestly, you can remove the Deposited from text and the Wallet button and just use the built in (<w3m-button />)[https://docs.walletconnect.com/appkit/react/core/components] button web component from walletconnectmodal

stormcloud266:
@youngkidwarrior updated with the web component

Screenshot 2024-07-25 at 3 00 51 PM

packages/app/features/deposit/web3/screen.tsx
```diff
@@ -127,6 +127,7 @@
   const { address: account } = useAccount()
```
youngkidwarrior, line 127: This flow is a little wonky. Remove the `useEffect` call. Just have people choose between clicking the Connect Button or directly depositing.

 `<w3m-button/>` should be able to cover all these cases. You might have to specify the preferred chain id. 

Also can you make sure the `<DepositAddress/>` component is visible before the wallet is connected. At the moment it only shows after connected

youngkidwarrior (changes requested):
looks good but lets cleanup this page a little, since we are already here

packages/app/features/deposit/web3/screen.tsx
stormcloud266: Not sure if i can pass a preferred chain id to the <w3m-button/> component, but it does force me to switch back to base if i change chains. 

I got those other two requests in as well. It shows when not connected as well

Screenshot 2024-07-26 at 2 30 37 PM

youngkidwarrior: Yeah you're almost there. The `Connect to Deposit` Button can just be replaced by <w3m-button>. It will handle the `isConnected` state and stuff itself

youngkidwarrior: Also button looks kinda funky in light mode

youngkidwarrior (approved):

PR #661, opened 2024-08-10
oauth2 authorization server

nicky-LV:


packages/app/package.json
0xBigBoss, line 1: shouldn't this go into the API? don't think it makes sense to include it in the app package. 

Somehow we either need to host this in our existing nextjs server or we can create a new app under our `/apps` directory for this server and deploy that separately.

supabase/migrations/20240810183922_create_oauth2_tables.sql
```diff
@@ -0,0 +1,34 @@
+-- adapted from: https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/resources/org/springframework/security/oauth2/server/authorization/client/oauth2-registered-client-schema.sql
+CREATE TABLE "public"."oauth2_clients" (
+    id SERIAL PRIMARY KEY,
+    client_id TEXT NOT NULL UNIQUE,
+    client_id_issued_at TIMESTAMPTZ DEFAULT NOW() NOT NULL,
+    client_name TEXT NOT NULL UNIQUE,
+    -- redirect_uri is stored in the table to prevent malicious URL redirect attacks
+    redirect_uri TEXT DEFAULT NULL,
```
0xBigBoss, line 8: should we make it an text[] column? so a client can support multiple redirect URIs? 

https://www.postgresql.org/docs/current/arrays.html

or does the oauth2 server recommend this approach? so if a client wants to have different envs for example, they need multiple clients?

nicky-LV: A client can have multiple redirect URIs (useful in different environments). Currently, each env has its own client ID.

On our side, I think allowing a client to have a singular `redirect_uri` would be easier. We only have to handle INSERTs.

Let's say we used `text[]`. Our client wants to add a new staging environment. We now need to handle UPDATEs, and this will probably come with a UI for them to edit the `redirect_uri`s they have registered.

Both approaches work though. Your thoughts on which route to go down?

packages/app/package.json
nicky-LV: Noted. Packaging in /packages/api

supabase/migrations/20240810183922_create_oauth2_tables.sql
0xBigBoss: Your approach of a single redirect URI per client ID is best I think, for now. I was just thinking of how google and/or facebook supports multiple per client ID (e.g. you want to test redirecting to your localhost environment).

Let's keep it simple as you suggested for v0

PR #665, opened 2024-08-12
Redirect user on invalid JWT for re-auth

nicky-LV:
Redirect user to /auth/sign-in on invalid JWT. Catches previously uncaught exception.

### DIFF
--- a/apps/next/utils/userProtected.ts
+++ b/apps/next/utils/userProtected.ts
@@ -4,6 +4,7 @@ import debug from 'debug'
 import type { GetServerSideProps, PreviewData } from 'next'
 import type { ParsedUrlQuery } from 'node:querystring'
 import { userOnboarded } from './userOnboarded'
+import { SUPABASE_SUBDOMAIN } from 'app/utils/supabase/admin'
 
 const log = debug('api:utils:userProtected')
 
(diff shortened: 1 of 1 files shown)


apps/next/utils/userProtected.ts
```diff
@@ -19,37 +19,49 @@ export function userProtectedGetSSP<
...
+      }
+      return getSSRResult
+    } catch (error) {
+      ctx.res.setHeader('Set-Cookie', 'sb-localhost-auth-token=; Max-Age=0; path=/')
       return {
         redirect: {
-          destination,
-          permanent: false,
+          destination: '/auth/sign-in',
+          permanent: true,
```
0xBigBoss, line 62:
```suggestion
          permanent: false,
```

0xBigBoss (approved):

apps/next/utils/userProtected.ts
```diff
@@ -19,37 +19,49 @@ export function userProtectedGetSSP<
...
+        ? await getServerSideProps(ctx)
+        : { props: {} as Props }
+      if ('props' in getSSRResult) {
+        // add the initialSession to page's getServerSideProps
+        // biome-ignore lint/suspicious/noExplicitAny: any is required here
+        ;(getSSRResult.props as any).initialSession = session
+      }
+      return getSSRResult
+    } catch (error) {
+      ctx.res.setHeader('Set-Cookie', 'sb-localhost-auth-token=; Max-Age=0; path=/')
```
0xBigBoss, line 58: Actually just read this, it won’t work in production

0xBigBoss (changes requested):

apps/next/utils/userProtected.ts
0xBigBoss: We need to use the value based on the servers request hostname

0xBigBoss (approved):

PR #679, opened 2024-08-20
splash page designs

stormcloud266:
* Updated home screen design
* Opens passkeys directly without having to navigate to the sign in page

Screenshot 2024-08-22 at 4 19 19 PM

Screenshot 2024-08-22 at 4 18 55 PM

PR #681, opened 2024-08-24
remove sign-in page

stormcloud266:
Redirects to `/` instead of `auth/sign-in`
Directly pops passkeys on `sign-up` when clicking "sign in"
Removes sign-in page, and removes `AuthLayoutOG` that used `auth/sign-in` but the component itself is not used anywhere


stormcloud266:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/681?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#681** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/681?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#679** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/679?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`

<h2></h2>

packages/app/features/unknown/screen.tsx
```diff
@@ -7,7 +7,7 @@ export function UnknownScreen() {
       <YStack jc={'center'} alignItems="center" f={1} gap="$6">
         <H1>Not found.</H1>
         <H2>Send, Instant Payments.</H2>
-        <Link href="/auth/sign-in">
+        <Link href="/">
```
youngkidwarrior, line 10: Wrapping Button in Link is actually an anti-pattern. Let's take this opportunity to fix this and use the `<LinkableButton/>` component

side note, do you like the name `LinkableButton?` Any better suggestions?

youngkidwarrior (changes requested):

youngkidwarrior:
Let's stack a PR on top of this and deal with these errors 

image

packages/app/features/splash/screen.tsx
```diff
@@ -259,7 +253,17 @@ function AuthButtons() {
...
       alignSelf="center"
     >
-      <SignInButton setError={setSignInError} size="$4" w="$12" />
+      <SignInButton
+        setError={setSignInError}
+        size="$4"
+        w="$12"
+        isSigningIn={isSigningIn}
+        setIsSigningIn={setIsSigningIn}
+      >
```
youngkidwarrior, line 262: Not quite. Instead of passing the props let's just switch between two buttons
```suggestion
     isSigningIn ? 
     <SignInButton 
        size="$4"
        w="$12"
        disabled
        disabledStyle={..}>
      <SignInButton.Text>
        Signing In...
      </SignInButton.Text>
     </SignInButton> :
     <SignInButton
        size="$4"
        w="$12"
        onPress={handleSignIn}
      >...
```

youngkidwarrior (changes requested):

youngkidwarrior (changes requested):
Same as change below. No need to have the sign in logic in the actual button

youngkidwarrior (changes requested):
Same as below

youngkidwarrior (commented):
Looking good. One small change and we are golden

packages/app/features/splash/screen.tsx
stormcloud266: Not sure I get what you're saying here. The SignInButton is already responsible for all the sign in logic, the only thing that needs changing between it's two current uses is the text and button text styles

youngkidwarrior: The logic for signing in, like the mutation and and query param handlers shouldn’t be in the button. The button should just be styles and html. Let the parent component control all the logic.

stormcloud266: The button is currently used in multiple spots, so I can move the logic to a hook or whatever, but moving it to the parent means duplicating all that code in both spots

youngkidwarrior: It's okay to repeat code a little in this case. Buttons shouldn't be managing all that state. If you'd like to move the sign in logic into a hook as well, I think that's a smart idea. If we find ourselves rewriting the sign in stuff a lot we can always add the hook later

stormcloud266: Okay i'll copy pasta and then we can do a hook later if we need to

youngkidwarrior (approved):

youngkidwarrior:
### Merge activity

* **Aug 29, 8:06 PM EDT**: @youngkidwarrior started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/0xsend/sendapp/681).

PR #686, opened 2024-08-29
Create transfer temporal workflow

youngkidwarrior:


youngkidwarrior:
* **#688** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/688?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#687** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/687?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#686** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `dev`

<h2></h2>

 Join @youngkidwarrior and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

apps/workers/src/client.ts
```diff
@@ -1,11 +1,9 @@
 import { Connection, Client } from '@temporalio/client'
-import {
-  // WorkflowA, WorkflowB,
+import { DistributionsWorkflow, TransferWorkflow } from '@my/workflows/all-workflows'
+import type { UserOperation } from 'permissionless'
+import('json-bigint-patch')
```
0xBigBoss, line 4: interesting.... never seen this one. it's a pretty neat implementation.

apps/workers/src/client.ts
```diff
@@ -21,23 +19,54 @@ export async function runWorkflow(): Promise<void> {
...
+  console.log('Started handle', handle.workflowId)
+  // optional: wait for client result
+  const result = await handle.result()
+  console.log('result: ', result)
+
+  return result
 }
 
-runWorkflow().catch((err) => {
+runDistributionWorkflow().catch((err) => {
```
0xBigBoss, line 48: we can comment out. I don't think @beezybarg is going to use the temporal version of the distributor.

packages/workflows/src/transfer-workflow/supabase.ts
```diff
@@ -0,0 +1,10 @@
+import { hexToBytea } from 'app/utils/hexToBytea'
+import { supabaseAdmin } from 'app/utils/supabase/admin'
+
+export async function fetchTransfer(hash: `0x${string}`) {
```
0xBigBoss, line 4: just a nit, i'd make it something closer to the actual result like `isTransferIndexed` or `transferExists` since it's only going to do a count query.

apps/workers/src/client.ts
0xBigBoss: I'm on the fence whether we should keep this patch. On the one hand, it's pretty amazing that it works (this works for client & workers right??).

The temporal way to do this correctly is to use a data converter. If it's easy to swap over then we should probably do it this way, https://github.com/temporalio/samples-typescript/blob/692fd263925a32add2c14fe7d7f2edc28b71b4ce/ejson/README.md

packages/workflows/src/transfer-workflow/activities.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (!!sender && !isAddress(sender))
+    throw ApplicationFailure.nonRetryable('Invalid send account address')
+  if (!!to && !isAddress(to)) throw ApplicationFailure.nonRetryable('Invalid to address')
+  if (!token || !isAddress(token)) throw ApplicationFailure.nonRetryable('Invalid token address')
+  if (typeof parsedAmount !== 'bigint' || parsedAmount <= 0n)
+    throw ApplicationFailure.nonRetryable('Invalid amount')
+  if (typeof parsedNonce !== 'bigint' || parsedNonce < 0n)
+    throw ApplicationFailure.nonRetryable('Invalid nonce')
+  try {
+    const userOp = await generateTransferUserOp({
```
0xBigBoss, line 31: we do we generate the userop again? we should just send what the client sent.

packages/workflows/src/transfer-workflow/wagmi.ts
```diff
@@ -0,0 +1,107 @@
...
+  callGasLimit: 100000n,
+  verificationGasLimit: 550000n,
+  preVerificationGas: 70000n,
+  maxFeePerGas: 10000000n,
+  maxPriorityFeePerGas: 10000000n,
+  paymasterVerificationGasLimit: 150000n,
+  paymasterPostOpGasLimit: 100000n,
+}
+
+export function generateTransferUserOp({
```
0xBigBoss, line 35: we shuold be able to remove this

packages/workflows/src/transfer-workflow/workflow.ts
```diff
@@ -0,0 +1,28 @@
+import { proxyActivities, ApplicationFailure } from '@temporalio/workflow'
+import type { createTransferActivities } from './activities'
+import type { Hex } from 'viem'
+
+const { sendUserOpActivity, waitForTransactionReceiptActivity, fetchTransferActivity } =
+  proxyActivities<ReturnType<typeof createTransferActivities>>({
+    startToCloseTimeout: '30 seconds',
```
0xBigBoss, line 7: probably want to increase this a bit, maybe to like 30-45 seconds.

Then, we also want to decrease the default `validUntil` here to like 30 seconds. https://github.com/0xsend/sendapp/blob/cbecd0d94f42a1bc243ef58387dcacc06670d981/packages/app/utils/signUserOp.ts#L22-L23

Maybe even sooner like 12-15 seconds? We may need some experimentation here to get right.

apps/workers/src/worker.ts
```diff
@@ -30,7 +40,7 @@ async function run() {
   // https://github.com/temporalio/samples-typescript/tree/main/hello-world-mtls
 
   // Step 2: Start accepting tasks on the `monorepo` queue
-  await worker.run()
+  await transferWorker.run()
```
0xBigBoss, line 43: it's only a transfer worker for now, but to minimize overhead, we'd run all of our workflow activities from the same worker deployment

apps/workers/src/client.ts
youngkidwarrior: It doesn't work that well unfortunately because it converts it to a number or string. So then the bundler won't accept it. I'll look into data converter

apps/workers/src/client.ts
youngkidwarrior: kk

packages/workflows/src/transfer-workflow/activities.ts
youngkidwarrior: generating it because passing the signed userop wasn't working. Data converter should fix this

packages/workflows/src/transfer-workflow/supabase.ts
youngkidwarrior: Good point, it doesn't need to return the tx. Only that it exists.

apps/workers/src/client.ts
youngkidwarrior: Seems that data converter doesn't support bigint :/

Gonna have to find a workaround

0xBigBoss: Try plugging in https://www.npmjs.com/package/superjson

youngkidwarrior: I'll try that. It also seems like we might be able to write a custom data converter.

0xBigBoss: Yup, that's what i meant, instead of `ejson` we would use `superjson`. 

Ideally, i think'd we'd just go for [protobuf](https://protobufjs.github.io/protobuf.js/), but unsure how much you want to tackle right now.

PR #688, opened 2024-08-29
Call transfer.withUserOp trpc mutation from frontend

youngkidwarrior:

PR #687, opened 2024-08-29
New trpc route that starts transfer workflow

youngkidwarrior:

PR #712, opened 2024-09-11
Show pending transfers from temporal in UI

youngkidwarrior:


youngkidwarrior:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/712?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#712** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/712?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#695** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/695?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#707** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/707?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#687** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/687?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#688](https://github.com/0xsend/sendapp/pull/688) <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/688?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#696** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/696?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#686** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/686?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`

<h2></h2>

 Join @youngkidwarrior and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/app/features/home/TokenDetailsHistory.tsx
```diff
@@ -1,16 +1,40 @@
...
+import { useSendAccount } from 'app/utils/send-accounts'
 
 export const TokenDetailsHistory = ({ coin }: { coin: coins[number] }) => {
+  const { data: sendAccount } = useSendAccount()
+  const [hasPendingTransfers, setHasPendingTransfers] = useState<boolean | undefined>(undefined)
+
+  const { data: pendingTransfers } = api.transfer.getPending.useQuery(
+    { token: coin.token, sender: sendAccount?.address ?? '' },
+    {
+      refetchInterval: 500,
```
0xBigBoss, line 18: this is too fast, don't want to put too much load on the servers. 1s is fine.

packages/app/features/home/TokenDetailsHistory.tsx
```diff
@@ -1,16 +1,40 @@
...
+      enabled: hasPendingTransfers === undefined || hasPendingTransfers,
+    }
+  )
+
   const result = useTokenActivityFeed({
     pageSize: 10,
+    refetchInterval: 1000,
```
0xBigBoss, line 25: why do we need them both to on the re-fetch interval? don't need to increase our request rate this much.

packages/app/features/home/TokenDetailsHistory.tsx
0xBigBoss, line 1: My recommendation would be to move out the pending stuff from this component and move it into the `useTokenActivityFeed`. Then in there, just have the query function return both results can make a return an union type between a pending transfer or a activity feed.

Or even just rename the current `useTokenActivityFeed` query to something like `useConfirmedTokenActivityFeed`, and then wrap both the pending one and confirmed one in a `useQueries`.

```ts
export function useTokenActivityFeed(params) {
const result = usePendingTransfers()
const result2 = useConfirmedTokenActivityFeed(params)
const data = [
  ...result,
  ...result2
]
return {
/// isloading, error, data etc
}
}
```

Then in each specific use query we can massage the data like we need for the components. Instead of having to do that in the components.

0xBigBoss: It just be easier to query both in a single query function though so you only need to refetch one query hook. instead of having to manage both in a component.

packages/app/features/home/TokenDetailsHistory.tsx
youngkidwarrior: Yeah this is just a rough solution to get everything working. We could adjust the refetch intervals to reduce load.

But, unless we can build an activity from the userop, we need both lists refetching, so that everything gets updated.

I don't see a simple way to match the userop with its matching activity. Maybe there is something we can pass to temporal to help with this?

As of now, they only are in this "watch" mode when there are pending transaction transfers. We could only allow this mode to run for the first 30 seconds or so

0xBigBoss: > I don't see a simple way to match the userop with its matching activity. 

why do we need to match? you already have them showing differently. Just fetch, merge the lists (with pending transfers always on top) and handle rendering the pending transfers differently.

Not in favor of those other workarounds. It's ok if we don't have all the details of an activity on our v0.

youngkidwarrior: When the pending transfers reach the confirmed state, they are removed from the pendingTransfers list. We need some way to filter the confirmed transfers so we only show them if they are not in the activity feed

PR #864, opened 2024-11-14
Handle existing user trying to use OTP

musidlo:
API part of the ticket: https://3.basecamp.com/5743564/buckets/36023883/todos/8021528909


packages/api/src/routers/auth.ts
```diff
@@ -31,6 +31,25 @@ export const authRouter = createTRPCRouter({
...
         })
       }
+
+      const { data, error } = await supabaseAdmin
+        .rpc('profile_lookup', { lookup_type: 'phone', identifier: phone })
+        .maybeSingle()
+
+      if (data) {
+        throw new TRPCError({
+          code: 'UNAUTHORIZED',
```
youngkidwarrior, line 41: Do you think we should return a case that is more specific. We have to consume this error on our frontend. In the end it will navigate to a new page 

Note you will also need to call this function again if the user is backing up their device.

0xBigBoss: > Note you will also need to call this function again if the user is backing up their device.

not necessarily, the passkeys should be cross-platform which means the new device should allow prompting for any passkey (even those not on the device).

Yeah, the error may be too cryptic. Instead something like 'User exists, please sign-in.'

musidlo:
Hey @youngkidwarrior @0xBigBoss, I slightly refatored error handling in auth, it was confusing to me. Now errors are only thrown, and route returns something when no errors occured. Ealier this code sometimes throwed errors and sometimes returned them - it was confusing for me. LTM what you think

packages/api/src/routers/auth.ts
```diff
@@ -9,49 +9,73 @@ const log = debug('api:auth')
...
-              })
-            }
-            const errMessage = r.error?.message.toLowerCase()
-            log('signInWithOtp', { errMessage })
-            return r
-          })
-        return { error: result.error }
-      } catch (error) {
+
+      if (!bypassPhoneCheck) {
```
musidlo, line 43: way to bypass checking if phone already exists, it will be needed when user is backing up

packages/api/src/routers/auth.ts
```diff
@@ -9,49 +9,73 @@ const log = debug('api:auth')
...
+        if (data) {
+          return {
+            wasPhoneAlreadyUsed: true,
+          }
+        }
+      }
+
+      const { error } = await supabaseAdmin.auth
+        .signInWithOtp({ phone: `${countrycode}${phone}`, options: { captchaToken } })
+        .then(async (r) => {
```
musidlo, line 57: here, mixing awaits with .then or .catch makes me super dizzy, I would refactor this to use only awaits but Im a bit scared of this snippet tbh

packages/api/src/routers/auth.ts
```diff
@@ -9,49 +9,79 @@ const log = debug('api:auth')
...
+
+        const { data } = await supabaseAdmin
+          .rpc('profile_lookup', { lookup_type: 'phone', identifier: phone })
+          .maybeSingle()
+
+        if (data) {
+          log('phone is already used', { phone })
+
+          return {
+            wasPhoneAlreadyUsed: true,
```
youngkidwarrior, line 54: NIT: instead of returning a boolean, just return the phone number used. 

If you're going to return a Boolean here, you might as well just return null

packages/api/src/routers/auth.ts
```diff
@@ -9,49 +9,79 @@ const log = debug('api:auth')
...
-          message: error instanceof Error ? error.message : 'Unknown error',
+          message: error.message,
         })
       }
+
+      log('successfully signed up with otp', { phone })
+
+      return {
+        wasPhoneAlreadyUsed: false,
+      }
```
youngkidwarrior, line 85: Ah I see what you're doing. This feels funny to me, what do you plan on using it for?

packages/api/src/routers/auth.ts
youngkidwarrior: I also wasn't a fan at first but I've come to rather like this syntax

Give it a chance

youngkidwarrior (approved):
Looks good

Just a NIT or two. Not a fan of returning booleans from API.

Approved, but don't merge this yet. Checkout your next branch off of this one

packages/api/src/routers/auth.ts
musidlo: So, that is how I plan to consume it

image

I want to make this straight forward to consume and use on client side, want to make returned schema look the same in all paths

packages/api/src/routers/auth.ts
youngkidwarrior: I mean that if we want to send the OTP, we will call this again, so we will need to bypass phone check

packages/api/src/routers/auth.ts
musidlo: yeah, of course it should be an enum, Im bad, refactoring it to be enum

packages/api/src/routers/auth/types.ts
musidlo, line 1: needed to move this types to dedicated file, to import it on UI, just like it is done with RecoveryOptions

0xBigBoss, line 1: not a big fan of separating out the types into their own file. but that's just my preference.

0xBigBoss: yeah i told the previous dev, Nicky, the same thing. It's all good, just letting you know my style preference. It's not a blocking change.

packages/api/src/routers/auth.ts
0xBigBoss: i'd rename it to be more clear, like bypassOnboardedCheck. Also not a big fan of this, but we need to keep it this way for now since we don't support any other way to sign in right now.

packages/api/src/routers/auth.ts
0xBigBoss: yeah it's a supabase choice to by default never throw exceptions and instead `return { data, error}`. It usually makes for much cleaner code, but this code is a hack to deal with local dev environments allowing us to not need to use a SMS provider like twilio.

PR #865, opened 2024-11-14
fix test, add downgrade verifier script, sync

0xBigBoss:


youngkidwarrior (approved):
Not sure about verifier stuff, everything else looks good

PR #882, opened 2024-11-22
Add X username to personal data

musidlo:


packages/app/components/FormFields/TextField.tsx
```diff
@@ -57,7 +58,7 @@ export const TextField = (
             color={props.labelProps?.color ?? '$olive'}
             {...props.labelProps}
           >
-            {label} {isOptional && '(Optional)'}
+            {label} {isOptional && !props.hideOptionalLabel && '(Optional)'}
```
0xBigBoss, line 61: I'd rather just remove the optional bit altogether, don't like this boolean style props that control components children rendering logic.

I don't think we have `(Optional)` anywhere in our designs anyway.

packages/app/features/account/settings/personal-info.tsx
```diff
@@ -7,89 +16,139 @@ import type { z } from 'zod'
...
+  const { mutateAsync: mutateAuthAsync } = useAuthUserMutation()
+  const { mutateAsync: mutateProfileAsync } = useProfileMutation()
+  const [formState, setFormState] = useState<FormState>(FormState.PersonalInfoForm)
+  const [errorMessage, setErrorMessage] = useState<string | null>(null)
+
+  function handleSuccessAuthUpdate() {
+    setFormState(FormState.VerificationCode)
+  }
+
+  async function handleSubmit() {
```
youngkidwarrior, line 43: I'd probably use two different forms, instead of grouping logic into one handleSubmit. It's more verbose but easier to grock

packages/app/utils/useAuthUserMutation.ts
```diff
@@ -6,6 +6,7 @@ import { useToastController } from '@my/ui'
 
 export const AuthUserSchema = z.object({
   phone: formFields.text.describe('Phone'),
+  xUsername: formFields.text.optional().describe('X'),
```
youngkidwarrior, line 9: Should probably be different forms since changing X doesn't require phone confirm

packages/app/utils/useUserPersonalDataMutation.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { AuthUserSchema } from 'app/utils/useAuthUserMutation'
+
+export const useProfileMutation = () => {
+  const supabase = useSupabase()
+  const { user } = useUser()
+  const queryClient = useQueryClient()
+  const toast = useToastController()
+
+  return useMutation({
+    async mutationFn(data: z.infer<typeof AuthUserSchema>) {
```
youngkidwarrior, line 15: Will need a different schema if we separate the forms

packages/app/features/account/settings/personal-info.tsx
musidlo: How do you imagine separating the forms from UI perspective? On the designs it's one form with submit button

youngkidwarrior: yeah thats an oversight of design. I guess we leave it like this and bring it up in the future

musidlo: from UX perspective I not if I would like to have two forms, one for phone only and the other for the rest, not sure there. this data won't be changed often, especially phone number

youngkidwarrior (approved):

youngkidwarrior (changes requested):
Make sure you pull the latest changes from dev. I think you are missing migrations

musidlo:
> Make sure you pull the latest changes from dev. I think you are missing migrations

done

youngkidwarrior (approved):

PR #900, opened 2024-11-26
Fix updating phone number

musidlo:


youngkidwarrior (commented):
Looking good, but I noticed that on dev and main, the existing phone number is not set as default value.

Can we update the form here to show the currently registered phone number as default value?

youngkidwarrior (changes requested):
Looking good, but I noticed that on dev and main, the existing phone number is not set as default value.

Can we update the form here to show the currently registered phone number as default value?

youngkidwarrior:
I'll double check. Thought I loaded it up on local and it didn't have it. I'll let you know by tomorrow

musidlo:
@youngkidwarrior did you had a moment to see it?

youngkidwarrior (approved):
lgtm

PR #944, opened 2024-12-11
Added profile info to transaction history

musidlo:
image

### DIFF
--- a/packages/app/features/profile/screen.tsx
+++ b/packages/app/features/profile/screen.tsx
@@ -65,7 +65,12 @@ export function ProfileScreen({ sendid: propSendid }: ProfileScreenProps) {
         f={1}
         gap={'$2'}
         display={isProfileInfoVisible ? 'none' : 'flex'}
-        $gtLg={{ display: 'flex', maxWidth: '50%', overflow: 'hidden', height: '80vh' }}
+        $gtLg={{
+          display: 'flex',
+          maxWidth: '50%',
+          overflow: 'hidden',
+          height: isWeb ? '80vh' : 'auto',
+        }}
       >
         <YStack
           f={1}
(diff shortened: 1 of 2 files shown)


packages/app/features/profile/screen.tsx
```diff
@@ -51,156 +55,193 @@ export function ProfileScreen({ sendid: propSendid }: ProfileScreenProps) {
...
-      ) : null}
-    </YStack>
+      {isProfileInfoVisible && (
+        <YStack
+          w={'100%'}
+          ai={'center'}
+          $gtLg={{
+            width: '35%',
+            minWidth: 400,
+            height: '81vh',
```
youngkidwarrior, line 226: Make sure we only use vh when we sure we are on the web.

Either use the `$platform-web` prop or import the `isWeb` flag

packages/app/features/profile/screen.tsx
```diff
@@ -267,3 +308,100 @@ const DatePill = ({ date }: { date: string }) => {
...
+                top={0}
+                left={0}
+                width="100%"
+                height="100%"
+                p={'$5'}
+                justifyContent="flex-end"
+                gap={'$3'}
+              >
+                <Paragraph size={'$9'} $theme-light={{ color: '$white' }}>
+                  Big investor
```
youngkidwarrior, line 367: ? Haven’t seen the designs but this doesn’t look right

youngkidwarrior (changes requested):
Just make sure we check for web when using vh

Make sure `Big Investor` is correct

packages/app/features/profile/screen.tsx
musidlo: lol

packages/app/features/profile/screen.tsx
musidlo: Im using vh only in $gtLg breakpoint, it wont apply for mobile

packages/app/features/profile/screen.tsx
musidlo: applied

packages/app/features/profile/screen.tsx
musidlo: discussed online, applied

youngkidwarrior (approved):

PR #961, opened 2024-12-19
Copy tweaks on transaction history

musidlo:


youngkidwarrior (changes requested):
Can you dim the background when modal is open, so we don't see two headers

image


NIT: I'm the type to change the routes to match the header for bookkeeping

send.app/profile becomes send.app/history

Your call

musidlo:
@youngkidwarrior did you look at proper code version? that's how it looks like right now
image
image

Im dont want to change route if not needed. People might have bookmarks or browsers history, also there will be automated tests to change, and finally I think that might change, a lot of stuff is changing right now

youngkidwarrior (approved):
LGTM.

if we want to link to this About modal, need to manage state in the url

PR #964, opened 2024-12-22
Move profile info state to url

musidlo:


packages/app/routers/params.tsx
```diff
@@ -8,6 +8,7 @@ export type RootParams = {
   nav?: 'home' | 'settings'
   token?: allCoins[number]['token']
   search?: string
+  isProfileInfoVisible?: boolean
```
youngkidwarrior, line 11: let's keep booleans out of the url. 

It's an antipattern to show more than one modal at a time. Let's just make a state similar to `nav`, that uses a string enum


```suggestion
  modal?: "profile"
```

youngkidwarrior (changes requested):

packages/app/routers/params.tsx
musidlo: applied

musidlo:
@youngkidwarrior rebased

PR #965, opened 2024-12-23
Temporal for transfers

youngkidwarrior:
Create transfer tempora workflow

New trpc route that starts transfer workflow

Encapsulate temporal into a package

Call transfer trpc mutation from frontend

Create transfer temporal workflow

Encapsulate temporal into a package

Create transfer tempora workflow

Encapsulate temporal into a package

Temporal cloud changes

remove debug env

Show pending transfers from temporal in UI


youngkidwarrior:
* **#965** <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/965?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/0xsend/sendapp/965?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

apps/next/next.config.js
```diff
@@ -38,6 +38,20 @@ const plugins = [
     },
     excludeReactNativeWebExports: ['Switch', 'ProgressBar', 'Picker', 'CheckBox', 'Touchable'],
   }),
+  (nextConfig) => {
+    return {
+      webpack: (webpackConfig, options) => {
+        // Add Temporal to externals when building for server
```
0xBigBoss, line 44: do you remember what this was avoiding? can we add to the comment?

packages/app/features/home/TokenDetailsHistory.tsx
```diff
@@ -1,93 +1,133 @@
...
+import { PendingTransferActivityRow, TokenActivityRow } from './TokenActivityRow'
+import { useSendAccount } from 'app/utils/send-accounts'
+import { zeroAddress } from 'viem'
 
 export const TokenDetailsHistory = ({ coin }: { coin: CoinWithBalance }) => {
-  const result = useTokenActivityFeed({
+  const { data: sendAccount } = useSendAccount()
+  const [hasPendingTransfers, setHasPendingTransfers] = useState<boolean | undefined>(true)
+  const { pendingTransfers, activityFeed } = useTokenActivityFeed({
+    address: sendAccount?.address ?? zeroAddress,
```
0xBigBoss, line 14: why did the API change so much on this?

0xBigBoss: I see, you are trying to convert the address param to the sender or from address? I'm confused why.

packages/app/features/home/utils/usePendingTransfers.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export function usePendingTransfers(params: {
+  address: Address
+  token: allCoins[number]['token']
+  refetchInterval?: number
+  enabled?: boolean
+}) {
+  const { address, token, refetchInterval, enabled } = params
+  return api.transfer.getPending.useQuery(
+    { token, sender: address },
```
0xBigBoss, line 16: we should not require the user to pass their address from the client, especially without validating it on the server.

environment.d.ts
```diff
@@ -16,6 +16,13 @@ declare global {
       NEXT_PUBLIC_SUPABASE_PROJECT_ID: string
       NEXT_PUBLIC_SUPABASE_GRAPHQL_URL: string
       NEXT_PUBLIC_MAINNET_RPC_URL: string
+      /**
+       * The URL of the ERC 4337 Account Abstraction Bundler RPC endpoint
+       */
+      BUNDLER_RPC_URL: string
+      /**
+       * The URL of the ERC 4337 Account Abstraction Bundler RPC endpoint
```
0xBigBoss, line 24: typo

packages/api/src/routers/transfer.ts
```diff
@@ -0,0 +1,108 @@
...
+import type { allCoins } from 'app/data/coins'
+
+const log = debug('api:transfer')
+
+export const transferRouter = createTRPCRouter({
+  withUserOp: protectedProcedure
+    .input(
+      z.object({
+        userOp: z.custom<UserOperation<'v0.7'>>(),
+        token: z.custom<allCoins[number]['token']>(), //@ todo: might be safer to decode the token from the userOp, to ensure we don't apply the wrong token
```
0xBigBoss, line 17: `//@ todo: might be safer to decode the token from the userOp, to ensure we don't apply the wrong token`

wdym?

packages/api/src/routers/transfer.ts
```diff
@@ -0,0 +1,108 @@
...
+        log(`Workflow Created: ${handle.workflowId}`)
+        return handle.workflowId
+      } catch (error) {
+        throw new TRPCError({
+          code: 'INTERNAL_SERVER_ERROR',
+          message: error instanceof Error ? error.message : 'Unknown error',
+        })
+      }
+    }),
+  getState: protectedProcedure.input(z.string()).query(async ({ input: workflowId }) => {
```
0xBigBoss, line 38: I don't see this used anywhere. can we remove?

packages/app/features/home/TokenActivityRow.tsx
```diff
@@ -109,3 +114,88 @@ export function TokenActivityRow({
     </XStack>
   )
 }
+
+export function PendingTransferActivityRow({
```
0xBigBoss, line 118: we should probably move all of this into when they are creating the workflow or within the workflows themselves.

packages/api/src/routers/transfer.ts
```diff
@@ -0,0 +1,108 @@
...
+import { TransferWorkflow, type transferState } from '@my/workflows/all-workflows'
+import type { allCoins } from 'app/data/coins'
+
+const log = debug('api:transfer')
+
+export const transferRouter = createTRPCRouter({
+  withUserOp: protectedProcedure
+    .input(
+      z.object({
+        userOp: z.custom<UserOperation<'v0.7'>>(),
```
0xBigBoss, line 16: it'd probably easier to accept more params here about the transfer to make it a lot easier on the client for showing pending data.

packages/api/src/routers/transfer.ts
```diff
@@ -0,0 +1,108 @@
...
+        }
+        return states
+      } catch (error) {
+        throw new TRPCError({
+          code: 'INTERNAL_SERVER_ERROR',
+          message: error instanceof Error ? error.message : 'Unknown error',
+        })
+      }
+    }),
+  getFailed: protectedProcedure
```
0xBigBoss, line 80: I don't see this used anywhere.

packages/app/features/home/TokenActivityRow.tsx
0xBigBoss: I'm talking about the parsing the token, amount, and the recipient.

packages/api/src/routers/transfer.ts
```diff
@@ -0,0 +1,108 @@
...
+        token: z.custom<allCoins[number]['token']>(),
+        sender: z.string(),
+      })
+    )
+    .query(async ({ input: { token, sender } }) => {
+      try {
+        const states: transferState[] = []
+        const client = await getTemporalClient()
+        const workflows = client.workflow.list({
+          query: `ExecutionStatus = "Running" AND WorkflowId BETWEEN "transfer-workflow-${token}-${sender}-" AND "transfer-workflow-${token}-${sender}-~"`,
```
0xBigBoss, line 63: do we need to worry about SQL injection here?

packages/app/features/send/confirm/screen.tsx
```diff
@@ -190,15 +179,21 @@ export function SendConfirm() {
...
+        userOp,
+        chainId,
         webauthnCreds,
+        entryPoint: entryPointAddress[chainId],
       })
-      assert(receipt.success, 'Failed to send user op')
-      setSentTxHash(receipt.receipt.transactionHash)
+      userOp.signature = signature
+
+      const workflowId = await transfer({ userOp, token: sendToken })
```
0xBigBoss, line 195: Reading things through now, the workflow ID should probably be `${userId}-${useropHash}`.

I think also, we should probably just create a new table in supabase to help us query this new data. So temporal workers inserts/updates into this new table with the temporal transfer workflows state. Then we can add triggers to this table to insert/update/delete the `activity` table.

That way it'll be easy to create, delete, and query them from the client side.

We can also create a new events in the activity table to help us display activities to the user. Or leverage the existing `SendAccountTransfersEventSchema` and extend it with a new event name and an optional state/status field.

This solution will allow us to show the transfer states easily and we can expand it to other workflows if we just use `data` column for the new temporal state table and are careful with our zod schemas.

packages/app/features/home/utils/usePendingTransfers.ts
youngkidwarrior: I’m in the middle of changing a lot of this flow. Good point though

packages/app/features/home/TokenActivityRow.tsx
youngkidwarrior: We have to show something in the frontend when the tx is pending

packages/api/src/routers/transfer.ts
youngkidwarrior: It will be

packages/api/src/routers/transfer.ts
youngkidwarrior: It’s can only be used for querying so not like they have any power.

packages/app/features/send/confirm/screen.tsx
youngkidwarrior: userOpHash sounds good. That fixes the nonce issue

I don’t agree with adding a new table. Why not just get the tx receipt back directly from temporal and use that in the frontend for optimistic updates. We don’t need all the different states and we don’t have to have 5 inserts per transaction. Isn’t this what temporal is for?

packages/app/features/home/TokenActivityRow.tsx
0xBigBoss: yeah, but the parsing and extracting this info can be done in the workflow.

packages/api/src/routers/transfer.ts
0xBigBoss: yah but we don't want to add SQL injection as a feature.

packages/app/features/send/confirm/screen.tsx
0xBigBoss: user op hash is a keccak hash of the user operation struct.

It's more about encapsulating this complexity of a workflow. The new table would allow us to remove the temporal trpc APIs, those are complicated and not going to scale well.

Using a new table and inserting rows from the workflow allows us to efficiently share data with the frontend and we can use existing libraries.

Then, the rows can trigger an insert into the activity table, call it `pending_send_account_transfer`, which looks similar but has new fields for tracking the state.

When a user op is finally success and we index the event using shovel, we update the row to success, then we can have an on update trigger to delete the `pending_send_account_transfer` row from the activity. no clean up needed.

we are evaluating spinning up our own, but still we shouldn't unnecessarily add API queries to it. It's main feature is workflow execution, and not querying data.

packages/app/features/home/utils/usePendingTransfers.ts
youngkidwarrior: Why does it matter though. Aren’t all these public transactions

packages/api/src/routers/transfer.ts
youngkidwarrior: I’ll make sure to sanitize when wet update the id

packages/app/features/send/confirm/screen.tsx
youngkidwarrior: I thought its main feature is managing the state of workflows and “queuing”. Wouldn’t that involve reading what is queued? Maybe I’m naive

0xBigBoss: yeah, but the user doesn't need to query temporal. let's try to limit querying/mutating the state of the temporal db within temporal workflows. We give users a view of their workflows by exposing it through supabase.

packages/app/features/home/utils/usePendingTransfers.ts
0xBigBoss: i don't see this as a good feature to have. being able to look up anyones pending transactions by their address. should be scoped to your own pending transfers.

youngkidwarrior: Also agree its not a good 'feature'. Just making a point that doing the work to obfuscate this stuff is kinda pointless no?

apps/next/next.config.js
youngkidwarrior: I don't, but it was something to do with cloud. I'll revisit

supabase/migrations/20250205035940_create_temporal_transfers_table.sql
```diff
@@ -0,0 +1,297 @@
...
+create policy "users can see their own temporal transfers"
+on "temporal"."send_account_transfers" as permissive
+for select to authenticated
+using (
+    user_id = auth.uid()
+);
+
+
+CREATE INDEX temporal_send_account_transfers_user_id_idx ON temporal.send_account_transfers(user_id);
+CREATE INDEX temporal_send_account_transfers_status_idx ON temporal.send_account_transfers(status);
```
0xBigBoss, line 44: think this one is overkill. don't think it will be a common query pattern. would rather see `created_at` index.

supabase/migrations/20250205035940_create_temporal_transfers_table.sql
```diff
@@ -0,0 +1,297 @@
...
+
+-- Create temporal schema
+CREATE SCHEMA IF NOT EXISTS temporal;
+
+-- Grant permissions for temporal schema
+GRANT USAGE ON SCHEMA temporal TO authenticated, service_role;
+
+-- Grant execute on functions to service_role
+GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA temporal TO service_role;
+ALTER DEFAULT PRIVILEGES IN SCHEMA temporal
```
0xBigBoss, line 11: Pretty sure service_role already has all of these permissions already. Likely the authenticated role is missing select queries.

packages/workflows/src/transfer-workflow/supabase.ts
```diff
@@ -0,0 +1,93 @@
...
+    .from('send_account_transfers')
+    .select('*', { count: 'exact', head: true })
+    .eq('tx_hash', hexToBytea(tx_hash))
+
+  log.info('isTokenTransferIndexed', { count, error, status, statusText })
+
+  return count !== null && count > 0
+}
+
+export async function isEthTransferIndexed(tx_hash: `0x${string}`) {
```
0xBigBoss, line 84: this won't work for when sending to EOA or contracts that don't emit a `Received` event

0xBigBoss: Likely, only thing to check for here would be that the user op is success. This approach would work if it was a transfer of eth between send accounts.

packages/workflows/src/transfer-workflow/workflow.ts
```diff
@@ -0,0 +1,35 @@
...
+const {
+  initializeTransferActivity,
+  sendUserOpActivity,
+  updateTemporalTransferSentStatusActivity,
+  waitForTransactionReceiptActivity,
+  isTransferIndexedActivity,
+  deleteTemporalTransferActivity,
+} = proxyActivities<ReturnType<typeof createTransferActivities>>({
+  // TODO: make this configurable
+  startToCloseTimeout: '45 seconds',
```
0xBigBoss, line 18: Is this the total time allowed for each activity?

packages/workflows/src/transfer-workflow/workflow.ts
```diff
@@ -0,0 +1,35 @@
...
+})
+
+export async function TransferWorkflow(userOp: UserOperation<'v0.7'>) {
```
0xBigBoss, line 21: where is it marked failed? not seeing it.

packages/workflows/src/transfer-workflow/workflow.ts
0xBigBoss: Think this setting should be a lot more generous, we want our workflow activities to re-run instead of timing out. Timing out makes me believe we will leave things in an inconsistent state. Imagine we have shovel downtime or base experiences a chain halt, temporal should just be able to pick up where we left off and continue, even if we pause the workers.

we can make 45 seconds for dev environments, but in prod it should be likely be much longer, idk let's start with like 10-15 minutes?

packages/app/utils/decodeTransferUserOp.ts
```diff
@@ -0,0 +1,23 @@
...
+import { sendAccountAbi, erc20Abi } from '@my/wagmi'
+import type { UserOperation } from 'permissionless'
+import type { allCoinsDict } from 'app/data/coins'
+
+export function decodeTransferUserOp({ userOp }: { userOp: UserOperation<'v0.7'> }) {
+  const { args } = decodeFunctionData({ abi: sendAccountAbi, data: userOp.callData })
+
+  const decodedTokenTransfer =
+    args?.[0]?.[0].data !== '0x'
+      ? decodeFunctionData({ abi: erc20Abi, data: args?.[0]?.[0].data })
```
0xBigBoss, line 11: what happens if someone passes a non erc20 transfer userop?

0xBigBoss (commented):
lookin pretty good. just some minor nits so far and some requested changes on the schema.

didn't review all the new changes in detail yet. will try to test it all later.

packages/workflows/src/transfer-workflow/workflow.ts
youngkidwarrior: It’s still a TODO

Just didn’t have anything left in the tank last night

packages/workflows/src/transfer-workflow/workflow.ts
youngkidwarrior: I think it will retry for 30 days

we can extend this time too for sure

supabase/migrations/20250205035940_create_temporal_transfers_table.sql
youngkidwarrior: should we let to_user also view temporal transfer activity? could show on both sides like a live feed

packages/temporal/build/payload-converter.cjs
```diff
@@ -0,0 +1,882 @@
...
+var isError = (payload) => payload instanceof Error;
+var isNaNValue = (payload) => typeof payload === "number" && isNaN(payload);
+var isPrimitive = (payload) => isBoolean(payload) || isNull(payload) || isUndefined(payload) || isNumber(payload) || isString(payload) || isSymbol(payload);
+var isBigint = (payload) => typeof payload === "bigint";
+var isInfinite = (payload) => payload === Infinity || payload === -Infinity;
+var isTypedArray = (payload) => ArrayBuffer.isView(payload) && !(payload instanceof DataView);
+var isURL = (payload) => payload instanceof URL;
+
+// ../../node_modules/superjson/dist/pathstringifier.js
+var escapeKey = (key) => key.replace(/\./g, "\\.");
```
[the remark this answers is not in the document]
youngkidwarrior: @0xBigBoss Any thoughts on this generated file?

packages/workflows/src/transfer-workflow/supabase.ts
youngkidwarrior: dropping "indexed" status anyway. Removing these

packages/app/features/home/utils/useTokenActivityFeed.ts
```diff
@@ -28,16 +30,24 @@ export function useTokenActivityFeed(params: {
 }): UseInfiniteQueryResult<InfiniteData<Activity[]>, PostgrestError | ZodError> {
   const { pageSize = 10, address, refetchInterval = 30_000, enabled = true } = params
   const supabase = useSupabase()
+  const { data: sendAccount } = useSendAccount()
+  const senderBytea = sendAccount?.address ? hexToBytea(sendAccount.address) : null
```
0xBigBoss, line 34: this shouldn't be needed. we should leverage row level security instead.

packages/app/features/send/confirm/screen.tsx
```diff
@@ -203,77 +185,44 @@ export function SendConfirm() {
         maxPriorityFeePerGas: feesPerGas.maxPriorityFeePerGas,
       }
 
-      log('gasEstimate', usdcFees)
-      log('feesPerGas', feesPerGas)
-      log('userOp', _userOp)
-      const receipt = await sendUserOp({
-        userOp: _userOp,
+      console.log('gasEstimate', usdcFees)
```
0xBigBoss, line 188: let's leave as debug log.

you can enable the logs in the window by setting `localStorage.debug = 'app:*'` and refreshing.

https://www.npmjs.com/package/debug#:~:text=hex%3A%2068656c6c6f20776f726c6421%20%2B0ms-,Browser%20Support,-You%20can%20build

packages/app/features/send/confirm/screen.tsx
```diff
@@ -203,77 +185,44 @@ export function SendConfirm() {
...
         webauthnCreds,
+        entryPoint,
       })
-      assert(receipt.success, 'Failed to send user op')
-      setSentTxHash(receipt.receipt.transactionHash)
+      userOp.signature = signature
+
+      const workflowId = await transfer({ userOp })
+
       if (selectedCoin?.token === 'eth') {
```
0xBigBoss, line 204: we can drop this now too, or do it async.

packages/app/features/send/confirm/screen.tsx
```diff
@@ -203,77 +185,44 @@ export function SendConfirm() {
...
+        router.replace({ pathname: '/', query: { token: sendToken } })
+      }
     } catch (e) {
       console.error(e)
       setError(e)
       await queryClient.invalidateQueries({ queryKey: [useAccountNonce.queryKey] })
     }
   }
 
-  useEffect(() => {
```
0xBigBoss, line 227: ahhh so clean now.

packages/app/utils/decodeTransferUserOp.ts
0xBigBoss: We're assuming the calldata are for the abi. we need to handle those errors in the caller and make a note of it in the jsdoc of this function.

packages/temporal/build/payload-converter.cjs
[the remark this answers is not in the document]
0xBigBoss: Hmm, likely this is fine. We should likely move this into a known generated code folder like `dist` maybe it'll ignore it then. This escapeKey is coming from superjson.

packages/app/utils/decodeTransferUserOp.ts
0xBigBoss: Screenshot 2025-02-14 at 07 30 55

packages/temporal/package.json
```diff
@@ -0,0 +1,38 @@
...
+    },
+    "./client": {
+      "types": "./src/client.ts",
+      "default": "./src/client.ts"
+    }
+  },
+  "scripts": {
+    "lint": "tsc",
+    "server": "temporal server start-dev --db-filename ./var/temporal.db",
+    "build": "esbuild --bundle --outfile=build/payload-converter.cjs --target=esnext --platform=node --external:@temporalio/common --external:@bufbuild/protobuf src/payload-converter.ts"
```
0xBigBoss, line 23: hmm. let's make sure those external deps are installed into the apps that depend on this package.

0xBigBoss: we should add those as peer deps on this package right?

supabase/migrations/20250205035940_create_temporal_transfers_table.sql
0xBigBoss: I don't think the to_user should see it the  temporal.send_account_transfers. They should see it in their activity feed already with the triggers you added below.

supabase/migrations/20250205035940_create_temporal_transfers_table.sql
```diff
@@ -0,0 +1,380 @@
...
+CREATE TRIGGER temporal_send_account_transfers_trigger_delete_activity
+  BEFORE DELETE ON temporal.send_account_transfers
+  FOR EACH ROW
+  EXECUTE FUNCTION temporal.temporal_send_account_transfers_trigger_delete_activity();
+
+
+-- When a send_account_transfer activity is inserted, delete any temporal_send_account_transfers
+-- with the same tx_hash from activity table.
+-- This prevents duplicate activities once a transfer is completed.
+create or replace function send_account_transfers_trigger_insert_activity() returns trigger
```
0xBigBoss, line 336: we gotta rethink this approach.

supabase/migrations/20250205035940_create_temporal_transfers_table.sql
```diff
@@ -0,0 +1,380 @@
...
+language plpgsql
+security definer as
+$$
+declare
+    _f_user_id uuid;
+    _t_user_id uuid;
+    _data jsonb;
+begin
+    -- Delete any temporal transfers with matching tx_hash
+    DELETE FROM activity a
```
0xBigBoss, line 346: this will cause us a lot of pain. It's going to lock too many rows.

Instead, select the workflow_id first from temporal table, then delete the rows matching. The following query would use the existing activity index (event_name, event_id).

```sql
delete from activity a
where event_name = 'temporal_send_account_transfers' and event_id in (
    select sat.workflow_id from temporal.send_account_transfers sat
    where extract(epoch from sat.created_at)::numeric < NEW.block_time;
);
```

Also, let's make it a new trigger on the `send_account_transfers` table instead of replacing the existing one. We need one for `send_account_receives` too.

Finally, please review the logic and ensure we're only deleting rows that really need to be deleted. How will we show users their transfers failed, this delete everything < block_time approach seems like we delete the failed ones too? Ideally, we don't want the receiving user to see failed transfers.

0xBigBoss: Otherwise, I am in favor of this approach. maybe we can add showing failed transfers as a follow up.

But we should definitely just make this a new trigger and function instead of replacing.

packages/app/utils/decodeTransferUserOp.ts
youngkidwarrior: if the calldata isn't erc20 compatiple, won't this error out? isn't that desirable behavior?

supabase/migrations/20250205035940_create_temporal_transfers_table.sql
youngkidwarrior: This comment is wrong. it doesn't use tx_hash, it uses block_time and created_at

my bad

youngkidwarrior (commented):
@0xBigBoss Which you prefer. one big migration or split up into multiple smaller migrations

supabase/migrations/20250205035940_create_temporal_transfers_table.sql
0xBigBoss: Function calls have very little overhead. The bulk of the work will still be the queries themselves in the functions so not much benefit of keeping them in a single trigger. IMO creating a new trigger is just a style and developer experience thing since they techinically are meant to do different things.

supabase/tests/temporal_transfers_test.sql
```diff
@@ -0,0 +1,213 @@
...
+    VALUES (
+        '\x1234'::bytea,
+        '\x5678'::bytea,
+        '123'::text
+    )
+    $$,
+    'Test activity update'
+);
+
+-- @TODO update this to test send_account_transfer insert
```
0xBigBoss, line 198: TODO

supabase/config.toml
```diff
@@ -5,8 +5,8 @@ project_id = "send"
 [api]
 enabled = true
 port = 54321
-schemas = ["public", "storage", "graphql_public"]
-extra_search_path = ["public", "extensions"]
+schemas = ["public", "storage", "graphql_public", "temporal"]
```
0xBigBoss, line 8: wonder if we need any changes in superbase dashboard

packages/workflows/tsconfig.json
```diff
@@ -10,18 +10,20 @@
       "app/*": ["../app/*"],
       "@my/wagmi": ["../wagmi/src"],
       "@my/wagmi/*": ["../wagmi/src/*"],
-      "@my/api/*": ["../api/src/*"]
+      "@my/workflows": ["./src/all-workflows.ts"],
```
0xBigBoss, line 13: why this one mapped to `all-workflows.ts`?

packages/workflows/package.json
```diff
@@ -4,26 +4,38 @@
...
+      "types": "./src/utils/index.ts",
+      "default": "./src/utils/index.ts"
+    },
     "./all-workflows": {
       "types": "./src/all-workflows.ts",
       "default": "./src/all-workflows.ts"
+    },
+    "./workflow-bundle": {
+      "types": "./workflow-bundle.d.ts",
+      "default": "./workflow-bundle.js"
```
0xBigBoss, line 23: i'd like to see it moved to a `dist` folder otherwise pretty sure linters and maybe other tools will start analyzing the generated code.

packages/workflows/tsconfig.json
youngkidwarrior: I think I was originally trying to have `@my/workflows` point to `all-workflows` so that the workflows were returned. I ended up not doing it though, this might be left over

packages/app/utils/decodeTransferUserOp.ts
0xBigBoss: That's what I meant, we should handle that here in the activity and throw non-retryable right? https://github.com/0xsend/sendapp/blob/2756ef4bcb9473fb02dbede8358006bb4452a80f/packages/workflows/src/transfer-workflow/activities.ts#L64-L65

I meant adding more context to the jsdoc for the function so devs can understand what it does and what it doesn't do.

packages/temporal/build/payload-converter.cjs
0xBigBoss, line 1: I don't think we commit the built file, it should be built during CI.

supabase/config.toml
youngkidwarrior: doesn't seem like it. Working in local

packages/temporal/package.json
youngkidwarrior: These?
{
   "@temporalio/client": "^1.10.1",
    "@temporalio/common": "^1.11.1",
    "superjson": "^2.2.1"
}

apps/next/styles/globals.css
```diff
@@ -5,7 +5,6 @@
 :root {
   height: 100%;
   background-color: transparent;
-  overflow: hidden;
```
0xBigBoss, line 8: what's this for?

packages/app/features/home/utils/useTokenActivityFeed.ts
```diff
@@ -80,7 +88,17 @@ export function useTokenActivityFeed(params: {
...
+    refetchInterval: ({ state: { data } }) => {
+      const { pages } = data ?? {}
+      if (!pages || !pages[0]) return refetchInterval
+      const activities = pages.flat()
+      const hasPendingTransfer = activities.some(
+        (a) =>
+          a.event_name === Events.TemporalSendAccountTransfers &&
+          !['cancelled', 'failed'].includes(a.data.status)
+      )
+      return hasPendingTransfer ? PENDING_TRANSFERS_INTERVAL : refetchInterval
```
0xBigBoss, line 100: nice.

packages/app/features/send/confirm/screen.tsx
```diff
@@ -207,16 +190,21 @@ export function SendConfirm() {
...
-      setSentTxHash(receipt.receipt.transactionHash)
-      if (selectedCoin?.token === 'eth') {
-        await ethQuery.refetch()
-      } else {
-        await tokensQuery.refetch()
+      userOp.signature = signature
+
+      const workflowId = await transfer({ userOp })
+
+      if (workflowId) {
```
0xBigBoss, line 206: are there cases where we don't get a workflowId? we should throw an error if we don't get one back.

packages/workflows/package.json
```diff
@@ -4,26 +4,37 @@
...
+      "default": "./dist/workflow-bundle.js"
     }
   },
-  "type": "module",
   "scripts": {
     "lint": "tsc",
-    "test": "jest"
+    "test": "jest",
+    "build": "yarn bundle",
+    "bundle": "yarn with-env node --import 'data:text/javascript,import { register } from \"node:module\"; import { pathToFileURL } from \"node:url\"; register(\"ts-node/esm\", pathToFileURL(\"./\"));' src/scripts/build-workflow-bundle.ts",
```
0xBigBoss, line 29: why do we need to build it `with-env`? what env vars did it need? I didn't see any in the script.

packages/workflows/src/transfer-workflow/activities.ts
```diff
@@ -0,0 +1,218 @@
...
+      try {
+        const bundlerReceipt = await waitForTransactionReceipt(hexHash)
+        if (!bundlerReceipt) {
+          throw ApplicationFailure.retryable('No receipt returned from waitForTransactionReceipt')
+        }
+        log.info('waitForTransactionReceiptActivity', {
+          bundlerReceipt: superjson.stringify(bundlerReceipt),
+        })
+        if (!bundlerReceipt.success) {
+          throw new Error('Transaction failed')
```
0xBigBoss, line 181: this should be non retryable.

supabase/migrations/20250307021828_create_temporal_transfers_table.sql
```diff
@@ -0,0 +1,391 @@
...
+GRANT ALL ON TABLE temporal.send_account_transfers TO service_role;
+
+alter table "temporal"."send_account_transfers"
+    enable row level security;
+
+create policy "users can see their own temporal transfers"
+on "temporal"."send_account_transfers" as permissive
+for select to authenticated
+using (
+    auth.uid() = user_id
```
0xBigBoss, line 43:
```suggestion
    (select auth.uid()) = user_id
```

Otherwise, auth.uid will be executed for every single row in the table.

supabase/migrations/20250307021828_create_temporal_transfers_table.sql
```diff
@@ -0,0 +1,391 @@
...
+);
+
+create policy "users can only see temporal transfers they initiated"
+on "public"."activity" as permissive
+for select to authenticated
+using (
+    (event_name != 'temporal_send_account_transfers') OR
+    (from_user_id = auth.uid())
```
0xBigBoss, line 51:
```suggestion
    (from_user_id = (select auth.uid()))
```

supabase/migrations/20250307021828_create_temporal_transfers_table.sql
```diff
@@ -0,0 +1,391 @@
...
+);
+
+CREATE INDEX temporal_send_account_transfers_user_id_idx ON temporal.send_account_transfers(user_id);
+CREATE INDEX temporal_send_account_transfers_created_at_idx ON temporal.send_account_transfers(created_at);
+CREATE UNIQUE INDEX temporal_send_account_transfers_workflow_id_idx ON temporal.send_account_transfers(workflow_id);
+
+CREATE OR REPLACE FUNCTION temporal.insert_temporal_token_send_account_transfer(
```
0xBigBoss, line 58: these functions are fine this time around, but if the logic is just a simple CRUD, we don't really need the functions, the client can pass everything we need and RLS will enforce our policies correctly.

youngkidwarrior: I was thinking we could do that. Just pass the whole data object every time

supabase/migrations/20250307021828_create_temporal_transfers_table.sql
```diff
@@ -0,0 +1,391 @@
...
+                                              ( select array_agg(name)
+                                                from tags
+                                                where user_id = to_p.id and status = 'confirmed' )
+           )::activity_feed_user end as to_user,
+       a.data                        as data
+from activity a
+         left join profiles from_p on a.from_user_id = from_p.id
+         left join profiles to_p on a.to_user_id = to_p.id
+where a.from_user_id = ( select auth.uid() )
+   or (a.to_user_id = ( select auth.uid() ) and a.event_name not like 'temporal_%')
```
0xBigBoss, line 389: this is so we don't show pending transfers to the recipient right?

supabase/migrations/20250307021828_create_temporal_transfers_table.sql
```diff
@@ -0,0 +1,391 @@
...
+    enable row level security;
+
+create policy "users can see their own temporal transfers"
+on "temporal"."send_account_transfers" as permissive
+for select to authenticated
+using (
+    auth.uid() = user_id
+);
+
+create policy "users can only see temporal transfers they initiated"
```
0xBigBoss, line 46: wait there are no RLS policies on activity table. let's drop this one.

Users query using the `activity_feed` view.

packages/workflows/src/transfer-workflow/activities.ts
youngkidwarrior: yes it gets caught and throws a nonRetryable, but first we set status = 'failed

packages/workflows/src/transfer-workflow/activities.ts
```diff
@@ -0,0 +1,218 @@
...
+        }
+        return bundlerReceipt.receipt
+      } catch (error) {
+        log.error('waitForTransactionReceiptActivity failed', { error })
+        const { error: updateError } = await updateTemporalSendAccountTransfer({
+          workflow_id: workflowId,
+          status: 'failed',
+        })
+        throw ApplicationFailure.nonRetryable(updateError?.message)
```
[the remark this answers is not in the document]
youngkidwarrior: lol

packages/app/features/send/confirm/screen.tsx
youngkidwarrior: we throw in the temporal api 
https://github.com/0xsend/sendapp/blob/d623de1d591e6954151683b69c362e0ae2a9ba86/packages/api/src/routers/temporal.ts#L27-L48

the catch should handle it. Though we could probably clean it up

supabase/migrations/20250307021828_create_temporal_transfers_table.sql
```diff
@@ -0,0 +1,276 @@
...
+  AFTER INSERT OR UPDATE ON temporal.send_account_transfers
+  FOR EACH ROW
+  EXECUTE FUNCTION temporal.temporal_send_account_transfers_trigger_activity();
+
+CREATE OR REPLACE FUNCTION temporal.temporal_send_account_transfers_trigger_delete_activity()
+  RETURNS TRIGGER
+  LANGUAGE plpgsql
+  SECURITY DEFINER
+  AS $$
+BEGINa
```
0xBigBoss, line 208: typo

supabase/migrations/20250307021828_create_temporal_transfers_table.sql
```diff
@@ -0,0 +1,276 @@
...
+    'sent',
+    'confirmed',
+    'failed',
+    'cancelled'
+);
+
+CREATE TABLE temporal.send_account_transfers(
+    id serial primary key,
+    workflow_id text not null,
+    user_id uuid not null DEFAULT uuid_nil(), -- rely on trigger to set user_id
```
0xBigBoss, line 26: we should not set a default then.

supabase/migrations/20250307021828_create_temporal_transfers_table.sql
```diff
@@ -0,0 +1,276 @@
...
+  ELSIF NEW.data ? 'sender' THEN
+    -- ETH transfer
+    SELECT user_id INTO _user_id
+    FROM send_accounts
+    WHERE address = concat('0x', encode((NEW.data->>'sender')::bytea, 'hex'))::citext;
+  END IF;
+
+  -- Validate user_id
+  IF _user_id IS NULL THEN
+    RAISE EXCEPTION 'No user found for the given address';
```
0xBigBoss, line 72: Hmmm. most likely not going to happen, but instead of blocking the write. Let's just raise a notice with the workflow ID and then return early.

supabase/migrations/20250307021828_create_temporal_transfers_table.sql
```diff
@@ -0,0 +1,276 @@
...
+for select to authenticated
+using (
+ (select auth.uid()) = user_id
+);
+
+CREATE INDEX temporal_send_account_transfers_user_id_idx ON temporal.send_account_transfers(user_id);
+CREATE INDEX temporal_send_account_transfers_created_at_idx ON temporal.send_account_transfers(created_at);
+CREATE UNIQUE INDEX temporal_send_account_transfers_workflow_id_idx ON temporal.send_account_transfers(workflow_id);
+
+CREATE OR REPLACE FUNCTION temporal.handle_transfer_upsert()
```
0xBigBoss, line 52: maybe a one liner comment explaining why this is a good idea or why we did it this way

packages/api/src/routers/temporal.ts
```diff
@@ -0,0 +1,73 @@
...
+            userOperation: userOp,
+            entryPoint,
+            chainId,
+          })
+          const { workflowId } = await client.workflow.start(TransferWorkflow, {
+            taskQueue: 'monorepo',
+            workflowId: `temporal/transfer/${user.id}/${userOpHash}`,
+            args: [userOp],
+          })
+          await baseMainnetClient.call({
```
0xBigBoss, line 42: Shouldn't the order be reversed? Simulate the call before creating the workflow?

packages/api/src/routers/temporal.ts
```diff
@@ -0,0 +1,73 @@
...
+            account: entryPointAddress[baseMainnetClient.chain.id],
+            to: userOp.sender,
+            data: userOp.callData,
+          })
+          log(`Workflow Created: ${workflowId}`)
+          const { data: transfer, error: transferError } = await supabaseAdmin
```
0xBigBoss, line 48: Let's add a withRetry? Viem has a good withRetry function.

packages/app/features/send/confirm/screen.tsx
```diff
@@ -13,44 +10,38 @@ import {
...
 import { useCoinFromSendTokenParam } from 'app/utils/useCoinFromTokenParam'
-import { allCoinsDict } from 'app/data/coins'
-import { IconCoin } from 'app/components/icons/IconCoin'
+import { allCoins, allCoinsDict } from 'app/data/coins'
 
 import debug from 'debug'
 import { useTokenPrices } from 'app/utils/useTokenPrices'
 
 const log = debug('app:features:send:confirm:screen')
+import { api } from 'app/utils/api'
```
0xBigBoss, line 40: nit: lets clean up the imports

packages/app/features/send/confirm/screen.tsx
```diff
@@ -111,13 +109,24 @@ export function SendConfirm() {
   } = useAccountNonce({
     sender: sendAccount?.address,
   })
-  const { data: userOp } = useGenerateTransferUserOp({
+
+  // Only generate UserOp when signing starts
```
0xBigBoss, line 113: comment does not make sense

packages/app/features/send/confirm/screen.tsx
```diff
@@ -111,13 +109,24 @@ export function SendConfirm() {
...
+  const { data: userOp, isLoading: isGeneratingUserOp } = useGenerateTransferUserOp({
```
0xBigBoss, line 114: We should use isPending instead of isLoading since isLoading will return the stale data.

packages/app/features/send/confirm/screen.tsx
youngkidwarrior: good catch

packages/api/src/routers/temporal.ts
youngkidwarrior: Better to not put retries in serverless functions imo, since they timeout

Instead we call it once and return state to the client. Then the client retries

youngkidwarrior: saving for future PR

youngkidwarrior:
### Merge activity

* **Mar 27, 4:57 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/0xsend/sendapp/965).

PR #980, opened 2025-01-02
deploy

0xBigBoss:
* fix(app): use submitted at for confirm wait time

this broke awhile back when we autowaited for the receipt.

* chore: fix distributions test, update wagmi

* chore: fix lint error

* chore: commit snaplet

* chore: fix unused variable

* chore: commit wagmi

* chore: fix e2e specs

* fix: redirect when eth send

add handle on unmount to prevent losing the pending send

* fix: only redirect if there is sentTxHash


youngkidwarrior:
Nice

PR #991, opened 2025-01-03
Redesign activity

musidlo:
image

### DIFF
--- a/packages/app/features/activity/screen.tsx
+++ b/packages/app/features/activity/screen.tsx
@@ -71,9 +71,6 @@ function ActivityBody() {
             opacity: 0,
             y: 10,
           }}
-          $gtLg={{
-            maxWidth: '50%',
-          }}
         >
           <RecentActivity />
         </YStack>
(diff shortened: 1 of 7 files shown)


packages/app/features/activity/screen.tsx
```diff
@@ -54,22 +64,17 @@ function ActivityBody() {
 
       {results === null && !isLoading && !error && (
```
musidlo, line 65: didnt use switch due to some part in the middle has to be on the screen all the time and looking for work around just to use switch is not worth it imo

packages/app/provider/tag-search/TagSearchProvider.tsx
```diff
@@ -81,6 +81,7 @@ export const TagSearchProvider = ({ children }: { children: React.ReactNode }) =
       setIsLoading(false)
       setError(null)
       setResults(null)
+      onSearch.cancel()
```
musidlo, line 84: added this due to some weird behaviour noticed, when deleting 2 last chars and doing it with small lag there might be a situation when state is clear but then overriden by search results that was queued already

0xBigBoss (approved):

packages/app/features/activity/screen.tsx
youngkidwarrior, line 1: Let's add a default state when there is no activity. Should just be a label that says "No Activity"

youngkidwarrior (approved):

PR #1003, opened 2025-01-09
Make transaction history more like a chat experience

musidlo:
https://github.com/user-attachments/assets/0cd4e33d-f120-449b-b035-01927d1588d4


youngkidwarrior:
Should probably hide send button when scrolling up

packages/app/features/profile/screen.tsx
```diff
@@ -58,26 +60,47 @@ export function ProfileScreen({ sendid: propSendid }: ProfileScreenProps) {
     setIsProfileInfoVisible((prevState) => !prevState)
   }
 
+  const handleLoadMore = async () => {
+    if (containerRef.current) {
+      setPreviousScrollHeight(containerRef.current.scrollHeight)
```
youngkidwarrior, line 65: Probably don’t need all this scroll mumbo jumbo. 

I believe there is a scrollview prop that tells you when you’re at the end.

packages/app/features/profile/screen.tsx
```diff
@@ -135,63 +180,44 @@ export function ProfileScreen({ sendid: propSendid }: ProfileScreenProps) {
...
-                                      currentUserProfile={currentUserProfile}
-                                    />
-                                  </Fade>
-                                </Fragment>
-                              )
-                            })
+
+                          const activities = (pages || [])
+                            .flatMap((activity) => activity)
+                            .sort((a, b) => a.created_at.getTime() - b.created_at.getTime())
```
youngkidwarrior, line 186: Let’s do this sort on the backend. should be able to pass `.order` to supabase

apps/next/pages/profile/[sendid].tsx
```diff
@@ -68,9 +67,7 @@ export const getServerSideProps = (async (ctx: GetServerSidePropsContext) => {
 }) satisfies GetServerSideProps
 
 Page.getLayout = (children) => (
-  <MobileButtonRowLayout.Profile>
```
Comment (youngkidwarrior) [inline apps/next/pages/profile/[sendid].tsx:71]: Why not use `MobileButtonRowLayout` here?

youngkidwarrior (changes requested):
Couple surfsce level comments. Mostly just want clarifications

packages/app/features/profile/screen.tsx
musidlo: I will give it a try

apps/next/pages/profile/[sendid].tsx
Comment (musidlo) [reply apps/next/pages/profile/[sendid].tsx:71]: This was working fine as whole page was scrollable and hide when scroll down. Now only a transactions container is scrollable - need to to that coz transactions container needs to be scrolled down at the beginning and at the same moment need to see button and header. Gonna give a try with scroll view, maybe it will work better

packages/app/features/profile/screen.tsx
musidlo: what I need is
- on init load scroll to the bottom of transactions container
- on load more, preserve scroll position, so user is in the same place as before clicking load more
scroll view does not support it any better than view, in either way it has to be handled manually

packages/app/features/profile/screen.tsx
musidlo: applied

apps/next/pages/profile/[sendid].tsx
Comment (musidlo) [reply apps/next/pages/profile/[sendid].tsx:71]: Let me describe the experience we aiming for when user approach this page:

- user can see profile header
- user can see transactions container scrolled to the bottom
- user can see send button
- transactions container takes full space between header and button, regardless device height

to fulfil this requirements this is what I did 

- make it possible to set height for container and main scroll view in HomeLayout so all the children can go with 100% height instead of trying to make best judgement with vh, which wont be the same for different size of mobile phones
- make this page take 100% space available 
- make transactions container to take 100% space available between profile header and send button
- make send button be part of the page layout

Why I dropped using this send button that is hiding when scrolling down?
Before newest transactions were on the top, right after profile header, whole page was scrollable, when user scroll down, send button would disappear, header would be out of screen and whole page is scrollable. That's good experience. But it can't be done like this when you want newest transactions to be on the bottom, because profile header would be out of the screen which is bad. And send button being not part of the page layout, would cover up the newest transaction, which also is bad. Introducing margin/padding to dodge that would result in weirdly cut scrolling area. My idea was to no longer scroll whole page, only transactions container.

I spend some time on many different variants of this, and this is the best experience I can do, it looks the same on all devices heights, button is aligned with amount and confirm send pages.

Comment (musidlo) [reply apps/next/pages/profile/[sendid].tsx:71]: image

Comment (musidlo) [reply apps/next/pages/profile/[sendid].tsx:71]: image

Comment (musidlo) [reply apps/next/pages/profile/[sendid].tsx:71]: image

youngkidwarrior (changes requested):
I'm getting some weird behavior with dates and scroll when I press load more. It's seems its always showing the oldest date.

I was actually write before with my comments about the list. Instead of using a `<YStack/>`, use a `<FlatList/>` from react native. It comes with all the props you'd need for this. 

Something like this pseudo code should work

```
 <FlatList inverted={true} onEndReached={handleLoadMore} data={pages}>
```

If you want to maintain access to tamagui styles, then you can use the `styled` function from `@my/ui`

```
const ActivityList = styled(FlatList)
```

packages/app/features/profile/screen.tsx
```diff
@@ -83,126 +61,99 @@ export function ProfileScreen({ sendid: propSendid }: ProfileScreenProps) {
...
+                        )
+                      }}
+                      onEndReached={fetchNextPage}
+                      ListFooterComponent={
+                        !isLoadingActivities && isFetchingNextPageActivities ? (
+                          <Spinner size="small" color={'$primary'} my={'$4'} />
+                        ) : null
+                      }
+                      inverted={true}
+                      showsVerticalScrollIndicator={false}
```
youngkidwarrior, line 149: nice

youngkidwarrior (approved):

PR #1014, opened 2025-01-14
Added activity details

musidlo:


packages/app/features/activity/ActivityAvatar.tsx
```diff
@@ -8,16 +8,36 @@ import {
   type Activity,
 } from 'app/utils/zod/activity'
 
-export function ActivityAvatar({ activity }: { activity: Activity }) {
+export function ActivityAvatar({
+  activity,
+  isDetailsView = false,
```
0xBigBoss, line 13: always try to avoid using boolean props to change the rendering logic of components. It doesn't scale well.

instead just include the Avatar component as part of the props type.

Something like this would be better IMO
```ts
export function ActivityAvatar({ activity, ...props }: Avatar & { activity: Activity }) {
```

then just forward the props.

musidlo: applied

packages/app/features/activity/ActivityDetails.tsx
```diff
@@ -0,0 +1,129 @@
...
+              />
+            </Stack>
+          </XStack>
+          <XStack gap={'$2'} ai={'center'}>
+            <Paragraph
+              size={'$9'}
+              fontWeight={700}
+              textTransform={'uppercase'}
+              // @ts-expect-error end value for text-align is valid value, left is not working properly in this case
+              ta={'end'}
```
0xBigBoss, line 91: Yeah it’s probably a quirk with react native and react native web. It’s probably working since it’s forwarding this directly to the html element. May not work on native. But that’s ok. Minor detail.

packages/app/utils/activity.ts
```diff
@@ -138,6 +138,42 @@ export function eventNameFromActivity(activity: Activity) {
   }
 }
 
+/**
+ * Returns the human-readable phrase for event name of the activity for activity details.
+ * @param activity
+ * @returns
+ */
+export function activityDetailsEventNameFromActivity(activity: Activity) {
```
0xBigBoss, line 146: I like the method. The name not so much. Too verbose. Not a big deal though. Would prefer something more succinct like phraseFromActivity or humanNameFromActivity

0xBigBoss (approved):

packages/app/features/activity/RecentActivity.tsx
```diff
@@ -1,30 +1,53 @@
...
-      gap="$5"
-      testID={'RecentActivity'}
-      $gtLg={{
-        maxWidth: '50%',
-      }}
-    >
-      <ActivityFeed {...result} />
-    </YStack>
+    <XStack w={'100%'} height={0} gap={'$5'} f={1}>
+      <ScrollView
```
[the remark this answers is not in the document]
musidlo: image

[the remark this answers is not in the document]
musidlo: left part of the page needs to be scrollable, we did this on few pages already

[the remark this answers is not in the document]
youngkidwarrior: Okay I trust you

PR #1023, opened 2025-01-20
bb/dev

0xBigBoss:
- **fix(shovel): add missing send token on base sepolia**
- **fix(next): use yarn install inline builds for debug**


apps/distributor/src/distributorv2.ts
```diff
@@ -327,12 +336,19 @@ export class DistributorV2Worker {
...
         const maxWeight = previousReward / BigInt(sendSlash.scaling_divisor)
-        acc[v.user_id] = {
+        const ceiling = {
           // Cap the weight to maxWeight
           weight: BigInt(v.weight || 0) > maxWeight ? maxWeight : BigInt(v.weight || 0),
           // @ts-expect-error @todo metadata is untyped but value is the convention
           ceiling: BigInt(v.metadata?.value || 0),
         }
+
+        if (distribution.number === 11) {
```
kolwaii-agent, line 346: NIT: Let's make this a const so that way we remember what `11` is, i.e PRE_SENDV0_MIGRATION_DISTRIBUTION

PR #1046, opened 2025-01-26
feat: add tranche_id to distributions

0xBigBoss:
This was a miss when deploying the new send token merkle drop for the send token upgrade.

The tranche starts at 3 since we reused the base merkle drop contract when migrating to Base.

We are going to use tranche ID 3 for the new send token merkle drop contract for distribution 11 after the send token upgrade.

### DIFF
--- a/apps/distributor/src/app.ts
+++ b/apps/distributor/src/app.ts
@@ -54,7 +54,7 @@ distributorRouter.post('/merkle', checkAuthorization, async (req: Request, res: 
   const { data: shares, error: sharesError } = await selectAll(
     supabaseAdmin
       .from('distribution_shares')
-      .select('index, address, amount', { count: 'exact' })
+      .select('index, address, amount_after_slash::text', { count: 'exact' })
       .eq('distribution_id', id)
       .order('index', { ascending: true })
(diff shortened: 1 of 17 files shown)


apps/distributor/src/app.ts
```diff
@@ -54,7 +54,7 @@ distributorRouter.post('/merkle', checkAuthorization, async (req: Request, res:
   const { data: shares, error: sharesError } = await selectAll(
     supabaseAdmin
       .from('distribution_shares')
-      .select('index, address, amount', { count: 'exact' })
+      .select('index, address, amount_after_slash::text', { count: 'exact' })
```
youngkidwarrior, line 57: Good catch

youngkidwarrior (approved):
Looks good. Would it not be better to redeploy the merkle drop contract, to avoid this added complexity?

youngkidwarrior (changes requested):
Getting reports that progress send_slash is not correct for larger holders

this is likely due to using amount instead of amount_after_slash somewhere

Let me look this over in closer detail and try to find where we’re going wrong.

PR #1085, opened 2025-02-16
feat: Send Earn

0xBigBoss:
TODO: Send Earn

- [x] Rewards activity
- [x] Earnings activity (deposits, withdrawals)
- [x] Claiming affiliate USDC rewards
- [x] Activity feed should show send earn deposit/withdraw instead of generic deposit/withdraw
- [x] Tests for claiming affiliate rewards
- [ ] Review
  - [x] TEMPORAL_ADDR
  - [x] TEMPORAL_NAMESPACE
  - [x] TEMPORAL_MTLS_TLS_CERT
  - [x] TEMPORAL_MTLS_TLS_KEY

deploy checklist:
- [x] deploy aabundler
- [x] contracts
  - [x] prep wallet with burn prefund
- [x] upgrade shovel
- [ ] deploy send app
- [ ] deploy workers
- [ ] deploy shovel
  -  [ ] update block number start


packages/app/features/earn/active/screen.tsx
```diff
@@ -0,0 +1,261 @@
...
+      },
+      // {
+      //   Icon: IconSendSingleLetter,
+      //   label: 'Rewards',
+      //   href: `/earn/${coin.data.symbol}/rewards`,
+      // },
+    ]
+  }, [coin])
+
+  if (!coin || !coin.data) {
```
0xBigBoss, line 48: Fix redirect bug

packages/app/features/earn/active/screen.tsx
```diff
@@ -0,0 +1,261 @@
...
+  const totalValue = useMemo(() => {
+    if (!balances.data) return '0'
+    if (!coin.data) return '0'
+    const totalAssets = balances.data.reduce((acc, balance) => {
+      return acc + balance.currentAssets
+    }, 0n)
+    return formatCoinAmount({ amount: totalAssets, coin: coin.data })
+  }, [balances.data, coin.data])
+
+  const isLoading = balances.isPending || coin.isPending
```
0xBigBoss, line 90: Use isLoading

supabase/migrations/20250312152453_send_earn.sql
0xBigBoss, line 1: Add a trigger creating a referral when a user deposits into a referral vault.

0xBigBoss:
Kolwaii, can you add a description and review this PR?

packages/app/features/activity/ActivityAvatar.tsx
```diff
@@ -62,28 +68,60 @@ export function ActivityAvatar({
...
+    }
+  }
+
+  if (isERC20Transfer) {
     // is transfer, but an unknown user
     const address = from_user?.id ? activity.data.t : activity.data.f
+    const name = addressBook?.data?.[address] ?? address
+
+    if (name === ContractLabels.SendEarn) {
+      if (from_user?.id) {
```
erick-kolwaii, line 93: What's difference between SendEarnEvent and an ERC20 transfer with ContractLabel of SendEarn?

packages/app/features/earn/deposit/hooks/index.ts
```diff
@@ -0,0 +1,231 @@
...
+ * Determine the vault to deposit into.
+ *
+ * Priority order:
+ * 1. If the user has existing deposits, use their current vault
+ * 2. If there's a referrer vault, use that
+ * 3. Otherwise, return null
+ *
+ * @returns A query result containing the vault address to deposit into
+ */
+export function useSendEarnDepositVault({
```
erick-kolwaii, line 72: So it's possible for the user to have multiple vaults on accident.

If the referer doesn't have a vault and the user deposits before the referer, this means the referer never get their downline?

packages/app/features/activity/ActivityAvatar.tsx
0xBigBoss: send earn event covers deposit and withdraw which are emitted from the vault contracts `emit Deposit` and `emit Withdraw`

They both originate from the same transaction onchain.

Right now we are indexing both, but we only show the erc20 transfers in the main activity feed. This is ActivityAvatar has to handle the three kinds of events though.

packages/app/features/earn/deposit/hooks/index.ts
0xBigBoss: > So it's possible for the user to have multiple vaults somehow (race condition or something)

if you were depositing in two different pages at the same time, yes, this could happen. Not really a big issue though, we can support multiple vaults in the future.

> If the referer doesn't have a vault and the user deposits before the referer, this means the referer never get their downline?

Yeah, if you weren't the referrer at the time of deposit, I don't see how you should get credit for deposits. Supporting multiple vaults fixes this, but not a high priority right now.

0xBigBoss: Though right now, the factory is the one handling the initial deposit so we can track your deposit vault and ensure the affiliate vault is created correctly. This reverts when you already have a deposit vault set so it would be the first one that lands onchain that wins.

PR #1107, opened 2025-02-19
Fix profile layout after safe area

youngkidwarrior:


youngkidwarrior (commented):
kolwaii can you summarize this PR for me

PR #1115, opened 2025-02-19
Disable send route when sending is locked

youngkidwarrior:


0xBigBoss (commented):
LGTM. My only gripe with the `useCoins` and `useCoin` hook is w don't provide an easy way for the caller to invalidate the queries so that we refetch. But that can be a follow up, things mostly work for the users cuz they refetch on mount.

PR #1118, opened 2025-02-23
Tags UX tweaks

musidlo:


itkunmi:
There should be an Device application for the Project to bring more people and there should be a permanent support team that if the users want to report anything there will be reply fastly

packages/app/features/account/sendtag/add/AddSendtagsForm.tsx
```diff
@@ -256,7 +256,7 @@ export const AddSendtagsForm = () => {
             size={'$5'}
             color={'$black'}
           >
-            confirm
+            continue
```
itkunmi, line 259: This is nice and good

packages/ui/src/components/LinkableButton.tsx
```diff
@@ -77,7 +77,7 @@ const LinkableButtonFrame = styled(Button, {
     disabled: {
       true: {
         pointerEvents: 'none',
-        opacity: 0.7,
+        opacity: 0.5,
```
itkunmi, line 80: I love the rate if this opacity

itkunmi (approved):

PR #1122, opened 2025-02-25
Added Don't Show Again On Deposit Dialog

SABITRADE:
I added the option to mark a checkbox ✅ to agree Once on the pop-up warning message cause it's annoying

### DIFF
--- a/packages/app/features/deposit/components/DepositAddress.tsx
+++ b/packages/app/features/deposit/components/DepositAddress.tsx
@@ -149,7 +149,9 @@ useEffect(() => {
         console.error('Failed to save dontShowAgain:', error)
       }
     }
-    setCopyAddressDialogIsOpen(true)
+    if (!dontShowAgain) {
+  setCopyAddressDialogIsOpen(true);
+    }
     return
   }
 
--- a/packages/app/features/deposit/components/DepositAddress.tsx
+++ b/packages/app/features/deposit/components/DepositAddress.tsx
@@ -61,14 +61,25 @@ function CopyAddressDialog({ isOpen, onClose, onConfirm }) { const [dontShowAgai
               3. I understand that if I make any mistakes, there is no way to recover the funds.
             </Paragraph>
 
-            <XStack justifyContent="flex-end" marginTop="$4" gap="$4">
-              <Dialog.Close asChild>
-                <Button br={'$2'}>Cancel</Button>
-              </Dialog.Close>
-              <Button theme="yellow_active" onPress={onConfirm} br={'$2'}>
-                <Button.Text col={'$color12'}>I Agree & Proceed</Button.Text>
-              </Button>
-            </XStack>
+            <XStack justifyContent="flex-end" marginTop="$4" gap="$4" ai="center">
+  {/* Checkbox on the left */}
+  <XStack ai="center" gap="$2">
+    <input
+      type="checkbox"
+      checked={dontShowAgain}
+      onChange={() => setDontShowAgain(!dontShowAgain)}
+    />
+    <Text fontSize="$3">Don't show again</Text>
+  </XStack>
+
+  {/* Buttons on the right */}
+  <Dialog.Close asChild>
+    <Button br={'$2'}>Cancel</Button>
+  </Dialog.Close>
+  <Button theme="yellow_active" onPress={onConfirm} br={'$2'}>
+    <Button.Text col={'$color12'}>I Agree & Proceed</Button.Text>
+  </Button>
+</XStack>
           </YStack>
         </Dialog.Content>
       </Dialog.Portal>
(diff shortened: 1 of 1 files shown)


youngkidwarrior:
we are going to need to use more than react state if we want this to persist across sessions

Use react native async storage so we persist the check in local storage

SABITRADE:
Should I proceed with using AsyncStorage for persistence? Is it already installed in the project, or would I need to add it?

youngkidwarrior:
@SABITRADE its already there. Here is an example that is used to persist the price hidden state

https://github.com/0xsend/sendapp/blob/107f35353ca5b6f358cca6897c4d72d327fc0c92/packages/app/features/home/TokenBalanceCard.tsx#L141C1-L141C6

itkunmi:
I love the new change on send.app about showing the dollar sign, and I think you should add many currency sign like euro  and many more sign.
So that will be good to. @youngkidwarrior 

Am Helloo from send africa

SABITRADE:
@youngkidwarrior you can now review again and tell me if you find anything wrong

SABITRADE:
I did a AsyncStorage so when a user checked once it stores it as a cache and when you are back again it checks the previous choice and refrain 😁

packages/app/features/deposit/components/DepositAddress.tsx
```diff
@@ -20,8 +20,9 @@ import { useState, useEffect } from 'react'
 import type { Address } from 'viem'
 import { IconCopy } from 'app/components/icons'
 import { useQRCode } from 'app/utils/useQRCode'
+import AsyncStorage from '@react-native-async-storage/async-storage'
 
-function CopyAddressDialog({ isOpen, onClose, onConfirm }) {
+function CopyAddressDialog({ isOpen, onClose, onConfirm }) { const [dontShowAgain, setDontShowAgain] = useState(false)
```
youngkidwarrior, line 25: can you please use our biome integration to format these changes

packages/app/features/deposit/components/DepositAddress.tsx
```diff
@@ -61,14 +62,33 @@ function CopyAddressDialog({ isOpen, onClose, onConfirm }) {
...
-                <Button.Text col={'$color12'}>I Agree & Proceed</Button.Text>
-              </Button>
-            </XStack>
+            <XStack justifyContent="flex-end" marginTop="$4" gap="$4" ai="center">
+  {/* Checkbox on the left */}
+  <XStack ai="center" gap="$2">
+    <input
+      type="checkbox"
+      checked={dontShowAgain}
+      onChange={async () => {
```
youngkidwarrior, line 71: don't do this inline. Pop this out into a new function

packages/app/features/deposit/components/DepositAddress.tsx
```diff
@@ -61,14 +62,33 @@ function CopyAddressDialog({ isOpen, onClose, onConfirm }) {
...
-                <Button br={'$2'}>Cancel</Button>
-              </Dialog.Close>
-              <Button theme="yellow_active" onPress={onConfirm} br={'$2'}>
-                <Button.Text col={'$color12'}>I Agree & Proceed</Button.Text>
-              </Button>
-            </XStack>
+            <XStack justifyContent="flex-end" marginTop="$4" gap="$4" ai="center">
+  {/* Checkbox on the left */}
+  <XStack ai="center" gap="$2">
+    <input
```
youngkidwarrior, line 68: We have a checkbox form component that has some helpful defaults. lets use that here

https://github.com/0xsend/sendapp/blob/db7163bd2fff181e55655bade3ee829bb4e9bee7/packages/app/components/FormFields/BooleanCheckboxField.tsx

packages/app/features/deposit/components/DepositAddress.tsx
youngkidwarrior, line 25: we won't need this local state if we are using the checkbox form and the async storage

packages/app/features/deposit/components/DepositAddress.tsx
```diff
@@ -81,6 +101,23 @@ export function DepositAddress({ address, ...props }: { address?: Address } & Bu
...
   const [copyAddressDialogIsOpen, setCopyAddressDialogIsOpen] = useState(false)
+  const [dontShowAgain, setDontShowAgain] = useState(false);
+
+useEffect(() => {
+  const loadDontShowAgain = async () => {
+    try {
+      const savedValue = await AsyncStorage.getItem('dontShowAgain');
+      if (savedValue !== null) {
+        setDontShowAgain(JSON.parse(savedValue));
+        setIsConfirmed(JSON.parse(savedValue)); // Auto-confirm if "Don't show again" was checked
```
youngkidwarrior, line 112: this isn't the right way to do this. Instead we should load this state from storage in the screen component and pass it as props to skip over the dialog entirely

https://github.com/0xsend/sendapp/blob/84472bd1e90b0a2fcb09d09ff8b6bce8052014f6/packages/app/features/deposit/crypto/screen.tsx

youngkidwarrior (changes requested):

youngkidwarrior:
@SABITRADE closing this due to inactivity

PR #1250, opened 2025-04-08
fix: correct logic for finding free key slot in backup creation process

0xBigBoss:


0xBigBoss:
Kolwaii, add a PR description.

PR #1331, opened 2025-04-16
Bring back send button on token activity feed

musidlo:


musidlo:
Hey Kolwaii, can you summarise this PR?

PR #1431, opened 2025-04-30
New onboarding

musidlo:
deployment checklist:

- [ ] supabase email sign up config
image
image
- [ ] make sure migration is last in the directory, if not rename it
- [ ] make sure Turnstile config is fine for given env (it had wrong key on vercel previous)


musidlo:
Hey Kolwaii, can you summarise this PR?

youngkidwarrior (dismissed):

packages/api/src/routers/tag/router.ts
```diff
@@ -12,11 +12,121 @@ import { byteaTxHash } from 'app/utils/zod'
...
+      })
+    }
+  }),
+  registerFirstSendtag: protectedProcedure
+    .input(SendtagSchema)
+    .mutation(async ({ ctx: { supabase, session, referralCode }, input: { name } }) => {
+      try {
+        const supabaseAdmin = createSupabaseAdminClient()
+
+        const { data: profile, error: profileError } = await supabase
```
youngkidwarrior, line 60: Potential speed improvements by turning this into a db function to avoid the round trips

packages/api/src/routers/tag/types.ts
```diff
@@ -0,0 +1,4 @@
+export enum SendtagAvailability {
+  Available = 'Available',
```
youngkidwarrior, line 2: Won't need this if we check availability from client

packages/app/components/FormFields/CountryCodeField.tsx
```diff
@@ -58,23 +80,17 @@ export const CountryCodeField = ({
           {...props}
         >
           <Select.Trigger
-            f={1}
```
youngkidwarrior, line 61: Future shit

Should really add a search here at some point

packages/app/features/account/sendtag/checkout/checkout-form.tsx
```diff
@@ -60,7 +60,7 @@ export const CheckoutForm = () => {
 /**
  * Shows the referral code and the user's profile if they have one
  */
-function ReferredBy() {
+export function ReferredBy() {
```
youngkidwarrior, line 63: export usually a good indicator to love to a new file

packages/app/features/account/sendtag/first/FirstSendtagScreen.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { YStack } from '@my/ui'
+import { useUser } from 'app/utils/useUser'
+import { FirstSendtagForm } from 'app/features/account/sendtag/first/FirstSendtagForm'
+
+export const FirstSendtagScreen = () => {
+  const { tags } = useUser()
+  const router = useRouter()
+
+  useEffect(() => {
+    if (tags && tags.length > 0) {
```
youngkidwarrior, line 12: Not a fan of this redirect pattern, but it works for now

packages/app/features/account/sendtag/first/FirstSendtagForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+      await registerFirstSendtagMutateAsync({ name })
+    } catch (error) {
+      form.setError('root', {
+        type: 'custom',
+        message: error.message,
+      })
+      return
+    }
+
+    await user?.updateProfile()
```
youngkidwarrior, line 50: Nice this should handle any stale queries

packages/app/features/auth/sign-up/screen.tsx
```diff
@@ -1,10 +1,364 @@
...
+    }
+  }, [user?.id, router.replace])
+
+  const createAccount = async () => {
+    const { data: sessionData } = await supabase.auth.getSession()
+    assert(!!sessionData?.session?.user?.id, 'No user id')
+    await createSendAccount({ user: sessionData.session.user, accountName: formName })
+  }
+
+  const handleSubmit = async ({ name }: z.infer<typeof SignUpScreenFormSchema>) => {
```
youngkidwarrior, line 94: This really where temporal would be helpful

packages/api/src/routers/tag/router.ts
```diff
@@ -12,11 +12,121 @@ import { byteaTxHash } from 'app/utils/zod'
...
+
+        const { error: insertError } = await supabaseAdmin
+          .from('tags')
+          .insert({ name, status: 'confirmed', user_id: session.user.id })
+
+        if (insertError) {
+          throw new Error(insertError.message || 'Unable to save tag')
+        }
+
+        if (referrerProfileLookupResult) {
```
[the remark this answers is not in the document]
musidlo: fetchReferrer uses `referrer_lookup` which uses `profile_lookup`. 
`profile_lookup` returns id only if quered user is the one who is querying

`case when p.id = ( select auth.uid() ) then p.id end              as id,`
src: `20241223152609_update_profile_lookup_public_phone_number.sql`

that being said, fetchReferrer never returns id of the referrer because
- it will return null as id if you try to look up any other profile than yours
- it will return null as the profile if you try to refer yourself
`    case !!new_referrer && new_referrer.refcode === profile.referral_code:`
src: `packages/app/utils/useReferrer.ts`

packages/api/src/routers/tag/types.ts
musidlo: answered here https://github.com/0xsend/sendapp/pull/1431#discussion_r2080219731

packages/app/components/FormFields/CountryCodeField.tsx
musidlo: if you start typing on desktop it scroll you to countries starting with letter you typed. But only desktop only 1st letter. Options part of this component is ass, agree we should rebuilt it, but its non blocking and low prio

musidlo: its not heavly used either

packages/app/features/auth/sign-up/screen.tsx
musidlo: agree

 I didnt use temporal coz I think ux is pretty solid, not bulletproof tho. it would take me more time to develop and as we move really with the flow lately this flow might change in a second, if we have we need temporal to improve ux and we have this new flow approved accepted liked whatever Im more than happy to add temporal to it

packages/app/features/account/sendtag/checkout/checkout-form.tsx
musidlo: what can I say, Im lazy mofo, applied

packages/api/src/routers/tag/router.ts
```diff
@@ -12,11 +12,121 @@ import { byteaTxHash } from 'app/utils/zod'
...
 import { z } from 'zod'
-import { createTRPCRouter, protectedProcedure } from '../trpc'
+import { createTRPCRouter, protectedProcedure, publicProcedure } from '../../trpc'
+import { SendtagSchema } from 'app/utils/zod/sendtag'
+import { SendtagAvailability } from './types'
 
 const log = debug('api:routers:tag')
 
 export const tagRouter = createTRPCRouter({
+  checkAvailability: publicProcedure.input(SendtagSchema).mutation(async ({ input: { name } }) => {
```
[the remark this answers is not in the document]
youngkidwarrior: youre right, we use server side props to fetch tag in [tag].tsx. my bad

packages/api/src/routers/tag/router.ts
0xBigBoss: Vic meant moving these 5 DB queries into a postgres function which would improve the latency pretty signficantly. And make this router code much simpler.

Not a big deal, probably still going to respond in under 2 seconds.

musidlo: Yeah I mean I got it, just dont see how latency between server and db would be a problem. Significant difference would be if it would be 1 query on db vs 5 queries, then yes, but I dont feel like it would be much difference between serwer doing 5 db queries and db function doing 5 quieries. Or am i wrong, server and db are not in the same machine? if you both talking to me about it, they probly not lol

packages/app/features/auth/sign-up/screen.tsx
```diff
@@ -1,10 +1,369 @@
...
+      await validateSendtag(name)
+
+      if (signUpFormState !== SignUpFormState.PasskeyCreationFailed) {
+        await signUpMutateAsync({
+          sendtag: name,
+          captchaToken,
+        })
+      }
+
+      setFirstSendtagCookie(name)
```
0xBigBoss, line 103: this isn't compatible with native.

packages/app/utils/send-accounts/useSignIn.ts
```diff
@@ -0,0 +1,41 @@
+import { api } from 'app/utils/api'
+import { signChallenge } from 'app/utils/signChallenge'
+import { bytesToHex, hexToBytes } from 'viem'
+import { RecoveryOptions } from '@my/api/src/routers/account-recovery/types'
+
+export const useSignIn = () => {
```
0xBigBoss, line 6: IMO this removes the benefits of using a mutation at all. Would be better to just create a new mutation that combines executes both mutations. Otherwise the caller has to handle the loading and error states themself.

packages/app/utils/tags/useValidateSendtag.ts
```diff
@@ -0,0 +1,23 @@
+import { SendtagSchema } from 'app/utils/zod/sendtag'
+import { SendtagAvailability } from '@my/api/src/routers/tag/types'
+import { api } from 'app/utils/api'
+
+export const useValidateSendtag = () => {
```
0xBigBoss, line 5: avoid homegrown hooks and leverage react query. loading and error states should always be handled.

packages/app/utils/useFirstSendtagCookie.ts
0xBigBoss, line 1: would be better to leverage react native async storage or something cross platform

supabase/migrations/20250507022536_query_user_by_phone.sql
```diff
@@ -0,0 +1,23 @@
...
+
+REVOKE EXECUTE ON FUNCTION query_webauthn_credentials_by_phone(phone_number TEXT) FROM PUBLIC;
+REVOKE EXECUTE ON FUNCTION query_webauthn_credentials_by_phone(phone_number TEXT) FROM anon;
+REVOKE EXECUTE ON FUNCTION query_webauthn_credentials_by_phone(phone_number TEXT) FROM authenticated;
+GRANT EXECUTE ON FUNCTION query_webauthn_credentials_by_phone(phone_number TEXT) TO service_role;
+
+-- Drop the existing trigger
+DROP TRIGGER IF EXISTS "insert_verification_tag_registration" ON "public"."tags";
+
+-- Recreate it for both INSERT and UPDATE
```
0xBigBoss, line 20: why we need it on update again?

musidlo: yes, I didnt want to alter confirm tags logic neither on server nor db. free tag doesnt have to be validated for receipt, transaction details and other stuff, also there will never be mixed free and paid in one action. didnt see any issues with inserting this free tag already confirmed. well, this trigger was, but I think it not too much if it runs also on insert

packages/app/features/auth/sign-up/screen.tsx
musidlo: changed to async storage

packages/api/src/routers/tag/router.ts
musidlo: added relation between tags and profiles tables to reduce api calls

packages/app/utils/send-accounts/useSignIn.ts
musidlo: changed to mutation

packages/app/utils/tags/useValidateSendtag.ts
musidlo: changed to mutation

PR #1455, opened 2025-05-07
native app, updates, fixes

0xBigBoss:
- **chore: use bun-overlay to support non-avx2 linux**
- **chore: update flake**
- **chore: refactor devShells in flake.nix for improved structure and macOS compatibility adjustments**
- **feat: clean up home layout**


packages/api/src/routers/temporal.ts
youngkidwarrior, line 1: On top of this, we should add a new route that shows the pending transaction
`/send/confirm/[workflowId]`

this should let us have more control over how we display the pending tx

packages/api/src/routers/temporal.ts
```diff
@@ -85,20 +86,50 @@ export const temporalRouter = createTRPCRouter({
...
+              code: 'PRECONDITION_FAILED',
+              message: 'Transfer not yet submitted',
+            })
           },
           {
-            retryCount: 10,
-            delay: 500,
+            retryCount: 20,
+            delay: ({ count, error }) => {
+              const backoff = 500 + Math.random() * 100 // add some randomness to the backoff
```
youngkidwarrior, line 129: nice

PR #1534, opened 2025-05-27
Top senders feed

musidlo:
## Summary 

The PR introduces a new feature to display top senders in the send screen, adding a new query to fetch top senders data and modifying the SendSuggestions component.

## Changes Made

- Added top senders query to fetch top senders data
- Modified SendSuggestions component to include top senders
- Created a new function useTopSenders to handle top senders query
- Added a new SQL function to calculate top senders based on send score

_written by Kolwaii, your beloved blockchain engineer AI agent_


0xBigBoss (approved):
Likely the query optimization isn't needed since our database is small, but it is a concern that the query optimizer may not know to how to build the best plan since I don't think we index `is_public` or `avatar_url` is not null which will end up always doing a sequential scan on the tags and profiles table since it's a left join.

musidlo:
I thought that its better to calculate send score only for filtered people but you might be right, gonna check it

musidlo:
@0xBigBoss actually it doesnt make a big of a difference, because you cannot limit before filtering by profiles. I did profiles filtering first then sendscore with limit, and the other way around, first send score then filtering with limit. I cannot do send score with limit, then filtering, coz Im gonna in the end less than 10 rows. These two variants are almost no diff.

PR #1934, opened 2025-09-20
Gate MV refresh with per-distribution advisory lock

youngkidwarrior:
Why:
Ensure only one transaction per previous distribution attempts the
refresh of private.send_scores_history. All other transactions skip MV
access entirely so other triggers are not delayed and commit latency is
minimized. This reduces redundant refresh work and limits commit-time
contention while keeping the logic inside the deferred trigger.

Test plan:
- Verify function compiles and trigger remains DEFERRABLE INITIALLY
  DEFERRED.
- Concurrency smoke: two sessions try
  SELECT pg_try_advisory_xact_lock(918273645, <prev_id>::int);
  Expect exactly one true; the other false.
- Insert transfers in a new distribution window under concurrency; check
  that only one commit runs the refresh (others skip), and no errors are
  raised.

### DIFF
--- a/docs/send-scores.md
+++ b/docs/send-scores.md
@@ -360,6 +360,52 @@ ORDER BY distribution_id DESC;
 - **Tag Search Weight**: `/1,000,000` divisor in ranking formula
 - **Access Levels**: RLS policies and role permissions
 
+## Automated refresh (deferrable trigger)
+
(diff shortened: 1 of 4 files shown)


youngkidwarrior:
### Merge activity

* **Sep 21, 5:44 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/0xsend/sendapp/1934).

